PR #69, opened 2022-02-10
Fix compilation errors for arch linux

FollowerOfBigboss:
Hi dail8859, i encountered your program while reading readme of Qt Advanced Docking System. I looked to your program and said "that's what i'm looking for". I fixed compilation errors for arch linux.

### DIFF
--- a/src/ads.pri
+++ b/src/ads.pri
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Notepad Next.  If not, see <https://www.gnu.org/licenses/>.
 
-QT += core gui widgets
+QT += core gui widgets gui-private
 
 windows {
 	# MinGW
--- a/src/NotepadNext/LanguageStylesModel.cpp
+++ b/src/NotepadNext/LanguageStylesModel.cpp
@@ -117,14 +117,14 @@ QVariant LanguageStylesModel::data(const QModelIndex &index, int role) const
             case 4: // Font
                 return editor->styleFont(index.row());
             case 5: // Size
-                return editor->styleSize(index.row());
+                return (int)editor->styleSize(index.row());
             case 6: // Fractional
-                return editor->styleSizeFractional(index.row());
+                return (int)editor->styleSizeFractional(index.row());
             case 7: // Bold
                 if (role == Qt::EditRole) return editor->styleBold(index.row());
                 else return editor->styleBold(index.row()) ? "True" : "False";
             case 8: // Weight
-                return editor->styleWeight(index.row());
+                return (int)editor->styleWeight(index.row());
             case 9: // Italic
                 if (role == Qt::EditRole) return editor->styleItalic(index.row());
                 else return editor->styleItalic(index.row()) ? "True" : "False";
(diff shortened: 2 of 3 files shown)


src/ads.pri
```diff
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Notepad Next.  If not, see <https://www.gnu.org/licenses/>.
 
-QT += core gui widgets
+QT += core gui widgets gui-private
```
dail8859, line 17: I'm not familiar with `gui-private. Does this hurt anything to keep this in for Windows builds? Or should there be a unix block that just adds `gui-private` to the unix builds?

dail8859:
Hi @FollowerOfBigboss!

Appreciate the PR and would love to see better Linux support :)

src/ads.pri
FollowerOfBigboss: Hi @dail8859 it doesn't hurt windows builds. It is required for docking code to compile. https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/blob/6fb45c055c25f838e7cdb9c6934c1f4968b2a1ea/src/ads_globals.cpp#L45 this line caused to issue.

FollowerOfBigboss:
I am downloading latest ubuntu iso right now. I will make some tests for ubuntu. After that it is good to merge.

dail8859:
> I am downloading latest ubuntu iso right now. I will make some tests for ubuntu. After than it is good for merge.

Awesome! I'll let you do whatever testing you need and merge when ready.

FollowerOfBigboss:
This is how NotepadNext looks like in Ubuntu 21.10
Screenshot_20220211_023808

This is how NotepadNext looks like in kde
Screenshot_20220211_025706


Build instructions for ubuntu 21.10
sudo apt install qtbase5-dev qt5-qmake qtbase5-dev-tools libqt5x11extras5-dev build-essential git
git clone https://github.com/dail8859/NotepadNext
cd NotepadNext
git submodule init
git submodule update
mkdir build
cd build
qmake ../src/NotepadNext.pro
make

Build will be in NotepadNext folder
to run build
./NotepadNext/NotepadNext

Weird but we have a issue in Ubuntu. It doesn't list files in file browser but drag and drop works. This issue doesn't exist in kde. It may gnome specific issue. I will investigate later.

Which looks like this.
Screenshot_20220211_023851

dail8859:
@FollowerOfBigboss 

Wow that is awesome!

> Build instructions for ubuntu 21.10

This is perfect!!! I had planned at one point to set up Ubuntu in a VM and document the process 👍 This will also help at some point to get it automatically building with Github Actions on Linux.

> Weird but we have a issue in Ubuntu. It doesn't list files in file browser but drag and drop works. This issue doesn't exist in kde. It may gnome specific issue. I will investigate later.

Yeah I'm sure there will be some oddities like this since its the first time it has been tested on different platforms.

dail8859:
> Weird but we have a issue in Ubuntu. It doesn't list files in file browser but drag and drop works.

I would guess it has to do with the use of [`QFileDialog::getOpenFileName()`](https://doc.qt.io/qt-5/qfiledialog.html#QFileDialog). The button on the bottom right of that screen would seem to indicate it is a Lua error (`attempt to index a nil value`). Internally Notepad Next uses Lua for its configuration and builds the file filters (e.g. *.txt, *.xml) so it still may be some internal bug.

dail8859:
I used Lua at the time because I was familiar with it and it gave me enough flexibility to focus on function rather than a specific implementation (plus I really hate XML 😄). I'm not overly attached to it for configuration but it works for now.

> Kde somehow tolerates it.

Yeah that is definitely weird. I need to pull out some of the embedded Lua code into actual files so that these types of issues could be debugged a bit easier.

dail8859:
FYI I Finally spun up an Ubuntu VM. Found I needed to also install `qtbase5-private-dev` for `gui-private`

FollowerOfBigboss:
Sorry for that i forgot to mention that.

PR #75, opened 2022-03-09
Build AppImage for Linux

dail8859:
This seems to be kind of the minimum changes needed to build an AppImage. The Github action completed successfully and I was able to download a build for Qt6 on an Ubuntu image that doesn't have Qt6 at all...so that's nice. I'm sure there's lots of room for improvement.


dail8859:
@FollowerOfBigboss

dail8859:
Things that still need addressed:

- [ ] Icons do not seem to work. Only the SVG is deployed with the App image. Probably need to include more sizes/formats. See: https://github.com/mapeditor/tiled/blob/316ceb4dfd5c5bb0ff26c4df885dd1c02d0c92b1/src/tiled/tiled.pro#L562-L604
- [ ] Not overly familiar with .desktop files so might need some improvements. E.g. https://github.com/pbek/QOwnNotes/blob/develop/src/PBE.QOwnNotes.desktop 
- [ ] Should the Notepad Next license be stored in the AppImage somewhere?
- [ ] Update `README.md`
- [ ] Update `doc/Building.md`

FollowerOfBigboss:
great work @dail8859 . It runs without a problem.

doc/Building.md
```diff
@@ -22,17 +24,31 @@ This document specifically describes how to build Notepad Next using Microsoft's
...
+## Building/Running Notepad Next
 
 1. Open `src/NotepadNext.pro` with Qt Creator
 1. Configure the project for 'Desktop Qt 5.15.2 MSVC2019 64bit'
 1. Press `Ctrl+R`
-1. Qt Creator will build and run the project.
\ No newline at end of file
+1. Qt Creator will build and run the project.
+
+# Linxu
```
FollowerOfBigboss, line 42:
```suggestion
# Linux
```

doc/Building.md
```diff
@@ -22,17 +24,31 @@ This document specifically describes how to build Notepad Next using Microsoft's
...
+Using a fresh Ubuntu 21.10 setup, the following script will install the needed dependencies and build the executable:
+
+```
+sudo apt install qtbase5-dev qt5-qmake qtbase5-dev-tools qtbase5-private-dev libqt5x11extras5-dev build-essential git
+git clone --recurse-submodules https://github.com/dail8859/NotepadNext.git
+cd NotepadNext
+mkdir build
+cd build
+qmake ../src/NotepadNext.pro
+make -j$(nprocs)
```
FollowerOfBigboss, line 53:
```suggestion
make -j$(nproc)
```

FollowerOfBigboss (changes requested):
Few corrections.

dail8859:
@FollowerOfBigboss 

> Few corrections.

Good catch! That's what I get for making these changes while being so tired 😊

dail8859:
I know there's still work to be done but I think at this point it would be best to just go ahead and get a release version of Linux out to test with and then work through the issues one by one as they come up.

Issue #86, opened 2022-04-11
About needs to be the last entry in the Help menu


tmssngr:
by an unwritten law.

dail8859:
Agreed, the Help menu grew a bit over time and needs reorganized.

mandyedi:
As a warm-up, I'd like to resolve this issue. How about this order:
Lua Console
Language Inspector
Editor Inspector
Check for Updates...
About Qt
About Notepad Next

dail8859:
@mandyedi 

> How about this order

That sounds reasonable.

PR #94, opened 2022-04-13
Fix Help menu action order

mandyedi:
This fixes the issue  **About needs to be the last entry in the Help menu #86**

### DIFF
--- a/src/NotepadNext/dialogs/MainWindow.cpp
+++ b/src/NotepadNext/dialogs/MainWindow.cpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * This file is part of Notepad Next.
  * Copyright 2019 Justin Dailey
  *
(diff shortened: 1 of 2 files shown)


src/NotepadNext/dialogs/MainWindow.ui
```diff
@@ -40,7 +40,7 @@
      <x>0</x>
      <y>0</y>
      <width>855</width>
-     <height>22</height>
+     <height>25</height>
```
dail8859, line 43: Was this change intentional? Thought I'm not sure this would even have any affect.

src/NotepadNext/dialogs/MainWindow.cpp
```diff
@@ -521,21 +521,22 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
...
+    ui->menuHelp->insertAction(editorInspectorDock->toggleViewAction(), languageInspectorDock->toggleViewAction());
 
-    EditorInspectorDock *editorInspectorDock = new EditorInspectorDock(this);
-    editorInspectorDock->hide();
-    addDockWidget(Qt::RightDockWidgetArea, editorInspectorDock);
-    ui->menuHelp->addAction(editorInspectorDock->toggleViewAction());
+    LuaConsoleDock *luaConsoleDock = new LuaConsoleDock(app->getLuaState(), this);
+    luaConsoleDock->hide();
+    addDockWidget(Qt::BottomDockWidgetArea, luaConsoleDock);
+    ui->menuHelp->insertAction(languageInspectorDock->toggleViewAction(), luaConsoleDock->toggleViewAction());
```
dail8859, line 539: Just a thought, would it be cleaner to use a single call to [insertActions()](https://doc.qt.io/qt-5/qwidget.html#insertActions) something like

```cpp
ui->menuHelp->insertActions(separator, { ___, ___, ___});
```

Haven't used that myself so not sure if there's some technical reason why it isn't viable.

dail8859 (commented):
Thanks for the PR! Left a few comments.

src/NotepadNext/dialogs/MainWindow.ui
mandyedi: It wasn't intentional. Some of the actions are added by the designer and when I change the order and remove one of the separators (it is put back by code), this height changes.

src/NotepadNext/dialogs/MainWindow.cpp
mandyedi: Do you mean constructing the action object while passing it to insertAction?
InsertActions takes a QAction as a second parameter, but we construct a LuaConsoleDock, which is a QDockWidget. After that hide() needs to be called on it, then we pass to insertAction using the toggleViewAction().
I don't think it could make it simpler or even possible.

dail8859: That's fine. I'll use the code as is. I'd rather have something working and clean it up later.

dail8859:
Thanks for the PR!

Issue #84, opened 2022-04-10
Flatpak packaging


abb128:
Flatpak offers some advantages over AppImage, namely automatic updates, more trustworthy distribution and sandboxing. It'd be nice to have a Flatpak package for those who prefer Flatpak.

zocker-160:
I am not the developer, but if there is interest, I would be willing to create the Flatpak.

zocker-160:
I managed to create a working flatpak, if anyone wants to give it a try:
[NotepadNext.zip](https://github.com/dail8859/NotepadNext/files/8465343/NotepadNext.zip)

so if @dail8859 is interested, we could push this to Flathub.

zocker-160:
@dail8859 let me answer your questions:

1. all flatpaks will be built on Flahub.org's servers fully automated once it is accepted (nothing needs to be done on your end in terms of building)
2. once the package is accepted, the submitter gets access to a repository which he can push directly to
see: https://github.com/flathub/flathub/wiki/App-Maintenance
and: https://github.com/flathub/flathub/wiki/App-Submission
3. it will be hosted on Flathub.org which has public built in statistics: 
see for example: https://beta.flathub.org/en/apps/details/org.mozilla.firefox

What I created above is a local Flatpak, which is meant for testing and it is not the "normal" way of distributing it.

> For now, I want to be very cautious what and how many release streams are supported because it can quickly get burdensome

I would recommend you to focus on Flatpak and Flatpak only, because it supports _all_ Linux distributions out there (unlike Appimage), it gives you great exposure to your application because all major distributions have flatpak support in the build in software store, so any user just needs to search for your application, presses install and done.

dail8859:
@zocker-160

Ah OK that makes **much** more sense now! I'm not overly set on sticking with AppImage, it was just easy and convenient to start with since I was looking for some way to get it deployed for Linux. I can definitely see some good advantages with flatpak (e.g. exposure and auto updates).

I'll need to do some more investigation with flatpak to see what all it entails. It seems you may have something semi-functional already.

dail8859:
@zocker-160 

v0.5 (bd6ca110d8219bde7467866e0d67da53353c69c8) has been released now.

zocker-160:
@dail8859 I just created the submission request to Flathub: https://github.com/flathub/flathub/pull/3024

zocker-160:
[...]

PR #97, opened 2022-04-15
add initial debian/ dir

bgstack15:
Intended for use on Debian-based systems. I built this on a Devuan Ceres (Debian Sid aka unstable) system. Anyone can use this as a base for building a package on similar platforms.


dail8859:
Thanks for the PR!

I'll need quite a bit more context on this PR exactly since I'm not a Debian user at all. There is a flatpak distribution in the works and I'm not sure if this will support Debian or not.

@zocker-160 can you comment on whether the flatpak will be available for Debian-based systems?

I want to be careful about how much maintenance burden there is with all the different build systems. Plus any additional burden of creating/testing/publishing a release. 

The current release is for Qt5 and this looks like Qt6. Its starting to introduce quite alot of variations in the releases which could make debugging and troubleshooting difficult.

bgstack15:
I know the build dependencies include some qt6 stuff. Those were necessary to build, probably because of qmake6. I don't know if qmake6 is directly related to qt6. Interestingly enough, the auto-resolved install-time dependencies were only:
```
 Depends: libc6 (>= 2.32), libgcc-s1 (>= 3.0), libqt5core5a (>= 5.15.1), libqt5gui5 (>= 5.14.1) | libqt5gui5-gles (>= 5.14.1), libqt5network5 (>= 5.0.2), libqt5printsupport5 (>= 5.0.2), libqt5widgets5 (>= 5.15.1), libstdc++6 (>= 11), libxcb1
```

zocker-160:
> can you comment on whether the flatpak will be available for Debian-based systems?

@dail8859 As explained in [my comment](https://github.com/dail8859/NotepadNext/issues/84#issuecomment-1095441039) in #84, Flatpak is supported by basically all major distributions out there, including Debian starting from Debian 8 (Jessie) from 2015.

see: https://en.wikipedia.org/wiki/Flatpak#Supported_out-of-the-box
and: https://www.flatpak.org/setup/

dail8859:
@zocker-160 

> Flatpak is supported by basically all major distributions out there

That's what I was thinking but wanted to make sure. 

@bgstack15

Thanks again for the PR but it seems flatpak will solve the need for a Debian specific distribution so I'll close this PR for now. If there's something I'm missing we can still discuss it, but there would have to be some major compelling reason to have something Debian specific since flatpak will cover nearly all major distributions.

Issue #99, opened 2022-04-16
Building for NixOS


SebTM:
Hey,

thanks for your great project! I've tried to package NotepadNext for Nix(OS) [https://nixos.org/]. For a build against a tag I would need a tag after #79, currently trying to build against master.

I come across errors I don't really know how to solve, I use qtbase-5.15.3 (which is one bugfix-level above the workflow), gcc/g++ 10.3.0:
```
[...]

PR #101, opened 2022-04-16
Reduce warnings

elcuco:
Contains several fixes to reduce warnings  - also https://github.com/dail8859/NotepadNext/issues/99.
After this PR - I only have warnings about unused functions, or variables.


dail8859:
Thanks for helping out with this!

PR #103, opened 2022-04-17
[i18n] init internationalization

inkydragon:
This is a POC (proof of concept).

Feel free to discuss details.
- Which folder should the translation files be placed in?
- Should we use the resource files?


src/NotepadNext.pro
```diff
@@ -23,14 +23,18 @@ TEMPLATE = subdirs
 
 SUBDIRS = NotepadNext
 
+TRANSLATIONS = \
+    ../i18n/NotepadNext.zh_CN.ts
+
```
inkydragon, line 28: The current issue is that the step to generate the qm file is not added to the qmake build process.

Possible fixes:
[qt - qmake and generated qm files - Stack Overflow](https://stackoverflow.com/questions/38588909/qmake-and-generated-qm-files)
```suggestion

LRELEASE_DIR = ../i18n
CONFIG += lrelease
```

But the above patch doesn't work for me.
The manually generated makefile does not have a target for generating qm files.

dail8859:
@inkydragon Thanks for the PR. These changes are pretty much what I was expecting which is good.

> Which folder should the translation files be placed in?

Currently I don't think it matters too much. It should be easy to change later.

> Should we use the resource files?

What are the options here? Embed the qm files in the application, or to distribute qm files individually with the exe?

inkydragon:
> Embed the qm files in the application, or to distribute qm files individually with the exe?

Yes, I prefer the latter.
People can easily generate and replace existing translations by themselves.

src/NotepadNext.pro
dail8859: I tried several combinations of settings but could not get qmake run lrelease to generate the qm files...oh well this is something that can be addressed later and just do it manually for now.

inkydragon: At least we can execute external commands in qmake through the `system` function.

[Automating generation of qm files - Qt Wiki](https://wiki.qt.io/Automating_generation_of_qm_files)

Something like:
```make
qtPrepareTool(LRELEASE, lrelease)
for(tsfile, TRANSLATIONS) {
    command = $$LRELEASE $$tsfile
    system($$command) | error("Failed to run: $$command")
}
``` 

If that doesn't work either, then do it manually for now.

dail8859: I somehow managed to get it to generate the qm files. I moved the `TRANSLATIONS` to `src/NotepadNext.pro` since it doesn't seem to like it in the SUBDIRS template.

dail8859:
@inkydragon

I wasn't able to push to your branch in your repo but I pushed the `i18n` branch to this repo. I added one more commit which I think fixes everything. The qm files are auto generated in the correct spot and they all get included in the zip and installer. Let me know what you think.

inkydragon:
> I wasn't able to push to your branch in your repo

That's true, but NOW maybe you can [commit changes to a pull request branch created from a fork - GitHub Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork).

image

I cherry-picked your patch, so we can continue to work on this pr.

dail8859:
Looks like I've got access now.

> I have checked the CI build product and it looks good.

Thanks for testing. What else needs done on this current PR then? I realize there's still translations to be done but as far as I'm aware things are working as expected.

inkydragon:
> What else needs done on this current PR then? 

I think this pr is ready for merge. But, I only tested the Windows version.

A quick look shows that there is no `.qm` file in `Appimage`.
So, the Linux and macOS version may need to be checked again.
Maybe need another pr.

> I realize there's still translations to be done but as far as I'm aware things are working as expected.

I'm working on Chinese translation, maybe this week or next week, new pr will come.

dail8859:
> So, the Linux and macOS version may need to be checked again.

Good point. I'm fine merging this as-is knowing that there's still improvements that can be made later. This way users that want to contribute new translations don't have to wait till everything is working perfectly. This is much better than no internationalization at all!

Thanks for all your work on this PR. This is a huge benefit to the project! 🎉

PR #104, opened 2022-04-17
Reminder that metainfo file also needs to include new releases

zocker-160:
because otherwise it will report the wrong application version in Flathub and Flatpak:
grafik

despite having v0.5 installed


dail8859:
So the metainfo should store a list of all past releases? Is the user able to install past version?

zocker-160:
The metainfo has to include at least the latest version, Past releases are optional, but I don't see a reason why to not include them. 

Imporant is, that the moment you create a tag, that the metainfo has the latest version in it, because that is the version that will be shown in Flathub.

> Is the user able to install past version?

Yes the user can revert / forward to whatever version he wants. Doing so does require commands in the terminal though, so I would guess that 99% of users will not do that.

dail8859:
Awesome. Thanks for the info and PR!

PR #106, opened 2022-04-19
add text/plain mime type to .desktop

zocker-160:
This will fix the issue of NotepadNext not being listed in the "open with" dialogue when opening text files. 
Currently it is not listed at all unless it is manually selected.


txtsd (approved):

dail8859:
Thanks!!

PR #108, opened 2022-04-19
fix missing icon in titlebar and taskbar

zocker-160:
_last PR for today, I hope I am not annoying with all those PRs :P_

I am not sure if this is a Linux only issue, since your Windows screenshot in the readme seems to have a titlebar icon, but on Linux there is only the default placeholder visible.

When adding the `windowIcon` property to the .ui file, it shows it correctly:
(left is current version, right is with this patch applied)
Screenshot_20220419_212544

same for the taskbar icon:
Screenshot_20220419_212643


dail8859:
> last PR for today, I hope I am not annoying with all those PRs :P

Not at all! I very much appreciate all the effort. Though I'm familiar with Linux and have used it in the past, its not my main day-to-day OS of choice so I'm glad to have someone willing to improve the Linux side of things!

> I am not sure if this is a Linux only issue

I've never noticed an issue on Windows. It has the title bar icon, task bar icon, etc. It doesn't have the big icon on the help menu but that might be Linux specific.

I want to poke around a bit on Windows before merging to see if I can understand why its Linux specific. Hopefully I'll get to it today 😄

dail8859:
Notes for myself for later. A quick search through the documentation indicates there's several ways to add an icon.

[QWidget::windowIcon()](https://doc.qt.io/qt-5/qwidget.html#windowIcon-prop) documentation:

> This property only makes sense for windows. If no icon has been set, windowIcon() returns the application icon QApplication::windowIcon()).

[QApplication::windowIcon()](https://doc.qt.io/qt-5/qapplication.html#windowIcon-prop) documentation:

> This property holds the default window icon

[QWindow::setIcon()](https://doc.qt.io/qt-5/qwindow.html#setIcon) documentation:

> The window icon might be used by the windowing system for example to decorate the window, and/or in the task switcher.

zocker-160:
I did a bit of testing in Qt and I think I know what is going on here.

If I understand your code correctly, at no point you are actually setting the icon, which means that Windows will use the icon inside the .exe as a fallback.
On Linux however the default Xserver window icon will be used as a fallback.

This also explains, why there is no icon in the about section on Windows, because there Qt uses the windows icon if set.

dail8859:
Yeah that sounds right. Since it worked on Windows I didn't give it another thought :)

dail8859:
> at no point you are actually setting the icon

Just confirmed if I use [QApplication::windowIcon()](https://doc.qt.io/qt-5/qapplication.html#windowIcon-prop) then it sets the icon for the entire application including showing up in the Help menu.

dail8859:
So I think the question is now, does it need to be a full 512x512 png?

It must be doing some massive down-scaling to fit it in a 16x16 icon for the windows title bar. However, everything looks fine...so maybe the concern has no merit.

image

zocker-160:
> So I think the question is now, does it need to be a full 512x512 png?

"need" no, but I think it makes sense, considering that:
- it could look bad in the about window
- many Linux distros (like ElementryOS) use large icons in their taskbar, so anything lower than 512x512 will not look great

both points are not really big deals though imo, if you are concerned, I think a 256x256 or 128x128 could also look "good enough".

zocker-160:
just curious: why do you prefer it "in code" vs the ui file?

dail8859:
> both points are not really big deals though imo

Since there's no known issue with 512x512 and the file size isn't a concern then we'll go with it for now. If there ever is an issue it can be addressed then. I tend to overthink things 😄 

> just curious: why do you prefer it "in code" vs the ui file?

The UI file is for the QMainWindow object, in code I set it on the QApplication object which as far as I know can't be configured with a UI file.

dail8859:
I took the opportunity to do a bit more cleanup, along with generating the PNG directly from the SVG file.

Just want to confirm the call to `setWindowIcon()` I added did fix the icon for Linux, correct?

zocker-160:
> Just want to confirm the call to setWindowIcon() I added did fix the icon for Linux, correct?

Negative in the current state it does not work. The reason is the name. I have uploaded `notepadnext.png` while in the code you reference `NotepadNext.png`

When setting the name correctly it works.

When you generate the png out of the svg, I think you can then remove the png I pushed.

dail8859:
> The reason is the name.

Ah good catch.

dail8859:
Thanks again for the PR!

Issue #107, opened 2022-04-19
Option to comment/uncomment code


Satharus:
This is more of a suggestion than an issue.

Like most text editors, a comment/uncomment feature should be present. It's usually Ctrl + / in most editors but Notepad++ used these:

![image](https://user-images.githubusercontent.com/29730368/164067402-6d44d4e3-b6f5-48c9-8122-bef0c845dd39.png)

dail8859:
Closed by #110

PR #110, opened 2022-04-19
Added the option to comment/uncomment lines of code.

Satharus:
Implemented the feature request in #107.

I set the shortcuts to Notepad++'s defaults. Although most text editors use Ctrl + / -which I also prefer- but I think since this is a reimplementation of Notepad++ we should try to stay as close to it as possible. Maybe we can make shortcuts configurable later on!

### DIFF
--- a/src/NotepadNext/NotepadNext.pro
+++ b/src/NotepadNext/NotepadNext.pro
@@ -78,9 +78,12 @@ SOURCES += \
     QuickFindWidget.cpp \
     RecentFilesListManager.cpp \
     SciIFaceTable.cpp \
+    ScintillaCommenter.cpp \
     ScintillaNext.cpp \
+    SelectionTracker.cpp \
     Settings.cpp \
     SpinBoxDelegate.cpp \
+    UndoAction.cpp \
     decorators/ApplicationDecorator.cpp \
     decorators/AutoCompletion.cpp \
     decorators/AutoIndentation.cpp \
@@ -128,9 +131,12 @@ HEADERS += \
     QuickFindWidget.h \
     RecentFilesListManager.h \
     SciIFaceTable.h \
+    ScintillaCommenter.h \
     ScintillaNext.h \
+    SelectionTracker.h \
     Settings.h \
     SpinBoxDelegate.h \
+    UndoAction.h \
     decorators/ApplicationDecorator.h \
     decorators/AutoCompletion.h \
     decorators/AutoIndentation.h \
--- a/src/NotepadNext/NotepadNextApplication.cpp
+++ b/src/NotepadNext/NotepadNextApplication.cpp
@@ -206,6 +206,8 @@ void NotepadNextApplication::setEditorLanguage(ScintillaNext *editor, const QStr
     const QString lexer = getLuaState()->executeAndReturn<QString>("return languages[languageName].lexer");
 
     editor->languageName = languageName;
+    editor->languageSingleLineComment = getLuaState()->executeAndReturn<QString>("return languages[languageName].singleLineComment");
+
     auto lexerInstance = CreateLexer(lexer.toLatin1().constData());
     editor->setILexer((sptr_t) lexerInstance);
 
--- a/src/NotepadNext/NotepadNextApplication.cpp
+++ b/src/NotepadNext/NotepadNextApplication.cpp
@@ -206,7 +206,7 @@ void NotepadNextApplication::setEditorLanguage(ScintillaNext *editor, const QStr
     const QString lexer = getLuaState()->executeAndReturn<QString>("return languages[languageName].lexer");
 
     editor->languageName = languageName;
-    editor->languageSingleLineComment = getLuaState()->executeAndReturn<QString>("return languages[languageName].singleLineComment or \"\"");
+    editor->languageSingleLineComment = getLuaState()->executeAndReturn<QString>("return languages[languageName].singleLineComment or \"\"").toUtf8();
 
     auto lexerInstance = CreateLexer(lexer.toLatin1().constData());
     editor->setILexer((sptr_t) lexerInstance);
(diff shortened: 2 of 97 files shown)


dail8859:
Thanks for the PR!

I haven't tried compiling running the code yet, but had some initial thoughts.

- I'd like to keep any language specific settings out of the MainWindow class. The MainWindow class is already larger than I'd like. Ideally these would go in each of the Lua files for each language. The grand plan is it should easy to add a new language and touch as few files as possible. (Notepad++ takes a modification of at least 10 files to add a new language).
- The actually commenting/uncommenting logic probably needs to go else where. Other parts of the code shouldn't be dependent on a MainWindow in order to trigger this behavior. Once you have an editor and a comment string, there's really nothing else you need...so this logic would probably be best put in the ScintillaNext class since it is very decoupled from the MainWindow.

> I think since this is a reimplementation of Notepad++ we should try to stay as close to it as possible

To be fair I'm not looking for a completely _pure_ reimplementation, so if there are changes that make sense, especially if the industry standard is different, now is the time to do them.

> Maybe we can make shortcuts configurable later on!

Definitely would be nice at some point!

Satharus:
Thanks for the feedback! Sorry for the late response I've been a bit busy the past 2 days and hadn't had the time to logon here.


> To be fair I'm not looking for a completely pure reimplementation, so if there are changes that make sense, especially if the industry standard is different, now is the time to do them.

Alright that totally makes sense I went ahead and changed the toggle comment to Ctrl+/ which is the default on most editors. I had to change the Reset Zoom shortcut to Ctrl+0 as it was originally set to Ctrl+/.

>  This logic would probably be best put in the ScintillaNext class since it is very decoupled from the MainWindow.

Done.

> Ideally these would go in each of the Lua files for each language.

Also done!

dail8859:
> Sorry for the late response

No problem at all. Everyone else has real lives outside of GitHub :)

Just glanced through the code real quick and that looks much better. When I get some time myself I'll go through the code in more detail.

Satharus:
Awesome! I am awaiting your review. 

Thanks!

dail8859:
I pushed some initial code changes.

- Remove need for `singleLineComment` in languages that don't have comments
- Removed explicit calls to `this->` and `ScintillaEdit::`
- Removed unneeded `singleLineCommentCharacters` member

A few things I noticed:
- This code only works for a single line, you can't select a range of lines like Notepad++
- Notepad++ uses a space after each character e.g. in C++ it comments a line starting with `// ` so I'm not sure if the code should add a space internally or if the `singleLineComment` should contain the space
- It would be nice to try to avoid creating a new `QString` for each line, especially if/when the code supports it, but I'd have to look at the code more to see if there's a better option.

If you feel like making the changes that's perfectly fine, if not I can take what you got and work from there.

Thanks

Satharus:
Awesome! Thanks for the cleanup. :smile: 

> This code only works for a single line, you can't select a range of lines like Notepad++

Well, yes, that's what I really had the time to implement for now, sadly. I am willing to contribute more to this project later on because I truly love the idea! I love notepad++ and how light it is and I was upset that it isn't available on Linux, since it is my main daily driver. So this project really is a Godsend hahahah

> Notepad++ uses a space after each character e.g. in C++ it comments a line starting with // so I'm not sure if the code should add a space internally or if the singleLineComment should contain the space.

I think we should add the space through the code, as the single line comment string doesn't actually contain the space as per each language's specifications. So I don't think it should be in the language file itself. That's just my opinion, though.

> It would be nice to try to avoid creating a new QString for each line, especially if/when the code supports it, but I'd have to look at the code more to see if there's a better option.

I guess there could be a way to look if the line has a comment string at the start using Scintilla functions without having creating a QString, but I'm not really sure what it could be tbh.

dail8859:
> Well, yes, that's what I really had the time to implement for now, sadly

Totally understandable.

> So I don't think it should be in the language file itself. That's just my opinion, though.

I guess I'd be more of the opinion to put the extra space in the `singleLineComment` variable that way a language could optionally include it or not. However since adding the space through code has worked well for the Notepad++ project then I guess it has never been an issue. So I'm fine doing it in code. Worst case scenario we find there is a particular language or something that shouldn't have the space and change it later.

> but I'm not really sure what it could be tbh.

I know there's way to find how far a line is indented so not sure if that could be used to see if the comment string is at that position. I haven't looked at the Notepad++ code to see how it does it there. I might be trying to prematurely optimize this section of code...I tend to overthink things.

The reason I'm a bit cautious about the QString() is because there could be cases of really long lines, or if the user selects *alot* of lines. Though I guess this would be an interesting case to profile to see just what kind of performance the two implementations would have...but if there's no other easy/obvious solution QString is fine until there's a reason to implement it otherwise.

Satharus:
> I guess I'd be more of the opinion to put the extra space in the singleLineComment variable that way a language could optionally include it or not.

Your project, your code! I think both options work just fine :smile: 

Thanks a lot for your efforts!

dail8859:
I was working with this code yesterday. I made some improvements but there's one issue that seems rather difficult to handle. When you insert the comment string, the selection is lost. Notepad++ handles this but the code is a bit of a mess to track the correct location.

https://github.com/notepad-plus-plus/notepad-plus-plus/blob/f6ade2c9157e7c6a56c95dfd04c436df512ca028/PowerEditor/src/Notepad_plus.cpp#L4980

Satharus:
> When you insert the comment string, the selection is lost.

I really didn't notice that when implementing it hahahaha ..

You're right though, I am not sure if we should mimic NP++'s code or if there is a smarter implementation :thinking:

dail8859:
Pushed a change which should handle the selection information. I very briefly tested this and it seems to give be similar behavior to Notepad++. Sure it needs more testing/tweaking.

dail8859:
The code isn't too messy. I don't like the repeated sections of code. I'll try to come up with a better way.

dail8859:
I refactored the code quite a bit. It is better than it was and seems to give correct behavior.

Satharus:
The code looks much better!

I gave it a test and it seems to be working perfectly fine. :muscle: :tada:

dail8859:
Apologies for the delay.

Thanks so much for the PR! Glad to see this functionality in the application. Best I can tell it works nearly identical to Notepad++ other than one minor difference...but I'll wait to see if anyone ever notices :)

Satharus:
Thanks for merging! Really happy to finally see some of my code become part of this project :smile:

Issue #56, opened 2021-12-28
Can you add simplified Chinese language


HeliusHao:
Can you add simplified Chinese language，thank

dail8859:
Hi @HeliusHao, thanks for opening this issue.

I'd be interested in working to get the application translated, however English is my only language plus I just don't know enough about the Qt translation system. 

If someone is interested in helping to lead this work to address some of the technical issues and the general process of translating the application, I'd be willing to help where possible.

lanmanck:
language can be added through community, I think.

mandyedi:
Hi @dail8859,
I'd like to work on adding the translation to NotepadNext, it seems quite an interesting task. I need to look up the technical background though, but meanwhile, I can add some UI for settings. I guess the change language menu should be in the preferences. I honestly don't know where it is in the original version, because I always used the English version.

dail8859:
Hi @mandyedi 

I think https://doc.qt.io/qt-5/internationalization.html would be the best place to start. 

> I can add some UI for settings

I don't believe that would be required to start with. Especially because I plan on changing the preferences dialog alot at some point. So if you do look at this don't put too much effort into making it look pretty.

The difficult part will be ensuring the translation is working and knowing how to do it. There's a snippet of code on that documentation that gets the user's Locale and attempts to load the appropriate translation if it exists.

mandyedi:
The translation process actually quite simple, but it might need some work to set up.
This is a very rough description of the main steps:
1. The strings meant for translation are marked using tr() in source code.
2. lupdate - A tool that scans the source files for tr() and places the strings in a .ts xml file.
3. A translator provides translations by opening the .ts file using Qt Linguist.
4. lrelease - A tool that takes the .ts which contains translated strings and converts them into a binary .qm format that can be loaded into an application at runtime.

I can write a detailed manual for developers and translators and add it to the docs folder.

dail8859:
That's good to hear. When I glanced at that documentation a while ago it seems to be pretty good about explaining everything, I just didn't have a need to go through it in detail at the time.

> I can write a detailed manual for developers and translators and add it to the docs folder.

Yeah that would be good. I don't think it needs to be overly detailed especially if the Qt documentation does a good job. But a list of steps would be good.

dail8859:
I fumbled through part of the process. No guarantee these translations are correct, but more just a test to see if it can load translations:

[...]

PR #113, opened 2022-04-20
Wrap string literals with tr() for translation

mandyedi:
I open this PR for the issue #56

For translation, the hardcoded string literals need to be wrapped with tr(). This does not affect the current state of the program, if no translation files are present, the default English phrases are displayed.
Note that in some cases I replaced QStringLiteral with tr, but it shouldn't be a problem at all. (in the function *void EditorInspectorDock::updateEditorInfo(ScintillaNext *editor)* and in several places in *EditorInfoStatusBar.cpp* file)

### DIFF
--- a/src/NotepadNext/dialogs/MainWindow.cpp
+++ b/src/NotepadNext/dialogs/MainWindow.cpp
@@ -271,7 +271,7 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
 
         QInputDialog d = QInputDialog(this);
         Qt::WindowFlags flags = d.windowFlags() & ~Qt::WindowContextHelpButtonHint;
-        int lineToGoTo = d.getInt(this, "Go to line", QString("Line Number (1 - %1)").arg(maxLine), currentLine, 1, maxLine, 1, &ok, flags);
+        int lineToGoTo = d.getInt(this, tr("Go to line"), tr("Line Number (1 - %1)").arg(maxLine), currentLine, 1, maxLine, 1, &ok, flags);
 
         if (ok) {
             editor->ensureVisible(lineToGoTo - 1);
@@ -507,8 +507,8 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
 
     ui->actionAboutNotepadNext->setShortcut(QKeySequence::HelpContents);
     connect(ui->actionAboutNotepadNext, &QAction::triggered, [=]() {
-        QMessageBox::about(this, "About Notepad Next",
-                            QString("<h3>Notepad Next v%1</h3>"
+        QMessageBox::about(this, tr("About Notepad Next"),
+                            tr("<h3>Notepad Next v%1</h3>"
                                     "<p>%2</p>"
                                     "<p>This program does stuff.</p>"
                                     R"(<p>This program is free software: you can redistribute it and/or modify
@@ -629,7 +629,7 @@ void MainWindow::newFile()
 
     static int count = 1;
 
-    app->getEditorManager()->createEmptyEditor(QString("New %1").arg(count++));
+    app->getEditorManager()->createEmptyEditor(tr("New %1").arg(count++));
 }
 
 // One unedited, new blank document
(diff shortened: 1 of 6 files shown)


dail8859:
@mandyedi Awesome! Thanks for going through and adding this in for the translations. I'll try to check out this PR later today but everything looks fine just glancing through it.

> Note that in some cases I replaced QStringLiteral with tr

Yeah that is totally fine since `tr()` takes a `const char *` rather than a `QString`

src/NotepadNext/dialogs/MainWindow.cpp
```diff
@@ -507,8 +507,8 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
 
     ui->actionAboutNotepadNext->setShortcut(QKeySequence::HelpContents);
     connect(ui->actionAboutNotepadNext, &QAction::triggered, [=]() {
-        QMessageBox::about(this, "About Notepad Next",
-                            QString("<h3>Notepad Next v%1</h3>"
+        QMessageBox::about(this, tr("About Notepad Next"),
+                            tr("<h3>Notepad Next v%1</h3>"
```
dail8859, line 511: This string should not be translated.

src/NotepadNext/dialogs/MainWindow.cpp
```diff
@@ -663,7 +663,7 @@ void MainWindow::openFileList(const QStringList &fileNames)
             QFileInfo fileInfo(filePath);
 
             if (!fileInfo.isFile()) {
-                auto reply = QMessageBox::question(this, "Create File", QString("<b>%1</b> does not exist. Do you want to create it?").arg(filePath));
+                auto reply = QMessageBox::question(this, tr("Create File"), QString(tr("<b>%1</b> does not exist. Do you want to create it?")).arg(filePath));
```
dail8859, line 666: QString() can be removed here:

```suggestion
                auto reply = QMessageBox::question(this, tr("Create File"), tr("<b>%1</b> does not exist. Do you want to create it?").arg(filePath));
```

src/NotepadNext/dialogs/MainWindow.cpp
```diff
@@ -1252,7 +1252,7 @@ void MainWindow::activateEditor(ScintillaNext *editor)
 void MainWindow::setLanguage(ScintillaNext *editor, const QString &languageName)
 {
     qInfo(Q_FUNC_INFO);
-    qInfo("%s", qUtf8Printable("Language Name: " + languageName));
+    qInfo("%s", qUtf8Printable(tr("Language Name: ") + languageName));
```
dail8859, line 1255: This is just for debugging and not needed to be translated.

dail8859 (changes requested):

src/NotepadNext/dialogs/MainWindow.cpp
mandyedi: What about Chinese and other languages that use different alphabets?

src/NotepadNext/dialogs/MainWindow.cpp
mandyedi: done

src/NotepadNext/dialogs/MainWindow.cpp
mandyedi: done

src/NotepadNext/dialogs/MainWindow.cpp
mandyedi: So actually my question is that why should we leave it in English? I misunderstood something, just forget the different alphabet. :)

dail8859: This call to `tr()` covers the entire block of text in the about dialog including the verbiage regarding the copyright ownership and the license itself.

I would not feel comfortable changing the legal terminology. The Free Software Foundation also has [no official translations](https://www.gnu.org/licenses/translations.en.html) of the license. So I feel its best to keep this in English for legal reasons.

dail8859:
I did some final cleanup and ran lupdate. Thanks for the PR!

src/NotepadNext/dialogs/MainWindow.cpp
mandyedi: I see, thank you for the clarification!

PR #116, opened 2022-04-21
fix developer name missing in metainfo

zocker-160:
fixes developer name missing on the Flathub page:
grafik


dail8859:
Would a full name be better than a username? Doesn't matter to me just didn't know if there was a best practice other projects use.

zocker-160:
> Would a full name be better than a username? Doesn't matter to me just didn't know if there was a best practice other projects use.

This is up to you, I think full name is what is supposed to be there, but I don't know if you want that. There are quite a few developers that prefer not using their real / full name.

So feel free to set whatever you want ;)

dail8859:
Guess full name (Justin Dailey) is fine.

dail8859:
Thanks for the PR!

PR #118, opened 2022-04-24
[i18n/zh-cn] Chinese translation

### LINKED ISSUE #56: Can you add simplified Chinese language

inkydragon:
Close #56

Most of the strings are already translated, but they need a second review.
I'll open the draft pr first to make it easier for people to review.

image

TODO:
- [ ] EditorlnspectorDock 27/35

Most of the text has already been translated.
For a few uncertain sentences, I have added `TODO` markers in the comments

### DIFF
--- a/i18n/NotepadNext.zh_CN.ts
+++ b/i18n/NotepadNext.zh_CN.ts
@@ -172,47 +172,47 @@
     <message>
         <location filename="../src/NotepadNext/widgets/EditorInfoStatusBar.cpp" line="98"/>
         <source>Length: %1    Lines: %2</source>
-        <translation type="unfinished"></translation>
+        <translation>长度: %1    行数: %2</translation>
     </message>
     <message>
         <location filename="../src/NotepadNext/widgets/EditorInfoStatusBar.cpp" line="109"/>
         <source>Sel: N/A</source>
-        <translation type="unfinished"></translation>
+        <translation>已选择: N/A</translation>
     </message>
     <message>
         <location filename="../src/NotepadNext/widgets/EditorInfoStatusBar.cpp" line="119"/>
         <source>Sel: %1 | %2</source>
-        <translation type="unfinished"></translation>
+        <translation>已选择: %1 | %2</translation>
     </message>
     <message>
         <location filename="../src/NotepadNext/widgets/EditorInfoStatusBar.cpp" line="125"/>
         <source>Ln: %1    Col: %2    </source>
-        <translation type="unfinished"></translation>
+        <translation>行: %1    列: %2    </translation>
     </message>
     <message>
         <location filename="../src/NotepadNext/widgets/EditorInfoStatusBar.cpp" line="143"/>
         <source>Macintosh (CR)</source>
-        <translation type="unfinished"></translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../src/NotepadNext/widgets/EditorInfoStatusBar.cpp" line="146"/>
         <source>Windows (CR LF)</source>
-        <translation type="unfinished"></translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../src/NotepadNext/widgets/EditorInfoStatusBar.cpp" line="149"/>
         <source>Unix (LF)</source>
-        <translation type="unfinished"></translation>
+        <translation></translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


dail8859:
Thanks for working on this!

I know there's been work to make sure strings are available for translation but if there's ways to improve the code to ease the work of translators then please let me know. When I started to develop the code I really didn't keep translation in mind. So there might be room for improvement since this is the first time it's being translated.

inkydragon:
> if there's ways to improve the code to ease the work of translators then please let me know.

Most of the code has no problems, a small amount of upstream code can be improved.

1. If multiple sentences will share a common part, then please split the shared part.
    e.g.: In Downloader: [QSimpleUpdater/Downloader.cpp](https://github.com/alex-spataru/QSimpleUpdater/blob/62e0ce7dde/src/Downloader.cpp#L212-L217)
    But this looks to be an upstream issue.

2. Use formatted string interpolation whenever possible. Because each string is separate in linguist, it is not easy to find the relevant context when translating if you concatenate the strings by hand

dail8859:
> But this looks to be an upstream issue.

Unfortunately true. 

> Use formatted string interpolation whenever possible.

Yeah this is what I had in mind when I was thinking about it. If there's cases that the code is doing weird string building/manipulation then I'm definitely willing to adjust code a bit if it is easier for translators.

mandyedi:
I have a small suggestion here, but maybe it is trivial anyway... Can we make changes, like formatted string interpolation, in a different PR? It would make it easier to start multiple translations, so if there is a major update regarding strings, we can just rebase our translation branch onto the master. I'd like to start Hungarian translation and if we change string literals in this PR, it would make it harder to maintain it. What are your opinions?

dail8859:
> Can we make changes, like formatted string interpolation, in a different PR?

Yeah that's fine with me. As you and @inkydragon work through translating this application for the first time there might be some changes needing made to the code, but hopefully things will get smoother with the translation process.

And @mandyedi I have not forgotten #98. Once we smooth out the translation process then I'll take a better look at that document so that other users can translate the application easier.

inkydragon:
@dail8859 

https://github.com/dail8859/NotepadNext/blob/a8fd45c08bf0d6cac68810b26c3635797ccdf3ce/src/NotepadNext/docks/EditorInspectorDock.cpp#L36
https://github.com/dail8859/NotepadNext/blob/a8fd45c08bf0d6cac68810b26c3635797ccdf3ce/src/NotepadNext/docks/EditorInspectorDock.cpp#L55
https://github.com/dail8859/NotepadNext/blob/a8fd45c08bf0d6cac68810b26c3635797ccdf3ce/src/NotepadNext/dialogs/MainWindow.cpp#L416

The "Print Preview" window that `Ctrl+P` opens also has a lot of text that needs to be translated, but I don't know where they come from.

image

----

Some thoughts on the use of words: "Wra&p Around" in "FindReplaceDialog".

Perhaps the word "cycling search" would be easier to understand?
The word "warp" I can only associate with auto line warp.
It was only after googling that I understood the meaning of the word: [What does the "Wrap around" check box do in 'search', 'replace' and 'mark' tabs? | Notepad++ Community](https://community.notepad-plus-plus.org/topic/16308/what-does-the-wrap-around-check-box-do-in-search-replace-and-mark-tabs)

dail8859:
a60a9f2bef7a7bcc35a570f572652847719f641c allows the strings to be translated.

---

> The "Print Preview" window that Ctrl+P opens also has a lot of text that needs to be translated, but I don't know where they come from.

This is purely a Qt dialog. It looks like there's [additional translation files](https://doc.qt.io/qt-5/internationalization.html#available-catalogs) that are used from the pre-compiled Qt libraries that would need included in the distribution. I'll need to look into this more.

---

> Some thoughts on the use of words: "Wra&p Around" in "FindReplaceDialog".

I agree the "wrap" terminology is not the clearest, however I would be cautious to invent new terminology. If there's some other established text editors that do something different I'd be interested to see what they do. The other possibility is to flip the logic and use something like "Stop at End of File" which I would think provides a description that is quite clear...but I would need a strong argument to go against industry standard.

dail8859:
I saw your other PR but is this ready to merge now?

inkydragon:
@dail8859  YES, after leaving it for a few days, I checked the translation again today, and it should be ready to merge.

dail8859:
@inkydragon Thanks for the help working through getting the application translated for the first time! Hopefully the makes it easier for future translations.

PR #128, opened 2022-05-04
[i18n] load translations for Qt

inkydragon:
The translation of `Qt_base` for Simplified Chinese is not complete, so this patch does not actually copy any files.

You can test the `windeployqt` process by adding a new ts file to the i18n folder.
eg.: 
```sh
cd i18n/
cp NotepadNext.zh_CN.ts  NotepadNext.zh_TW.ts
cp NotepadNext.zh_CN.ts  NotepadNext.en.ts
```

After `jom package`, you may get `build\package\translations\qt_*.qm`

### DIFF
--- a/src/NotepadNext.pro
+++ b/src/NotepadNext.pro
@@ -34,10 +34,22 @@ win32 {
         xcopy $$shell_path($${OUT_PWD}/NotepadNext/LICENSE) $$shell_path($${OUT_PWD}/package/) /Y && \
         xcopy $$shell_path($${OUT_PWD}/NotepadNext/i18n/*.qm) $$shell_path($${OUT_PWD}/package/i18n/) /Y &&
 
+    # Get all translation language names
+    #   get full path:      `../i18n/NotepadNext.en.ts`
+    TRANSLATION_LANGS = $$files($$PWD/../i18n/*.ts)
+    #   keep filename:      `NotepadNext.en.ts`
+    TRANSLATION_LANGS = $$basename(TRANSLATION_LANGS)
+    #   remove App name:    `en.ts`
+    TRANSLATION_LANGS = $$replace(TRANSLATION_LANGS, NotepadNext.,)
+    #   remove `.ts` ext:   `en`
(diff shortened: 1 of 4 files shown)


dail8859:
I looked at this real quick and this makes sense. I'll hold off merging this until #118 is merged. Let me know when #118 is ready.

dail8859:
A few initial thoughts:

- This will only work for Windows since it relies on `jom package` to call `windeployqt`
- Since the Qt translations will only be visible after its packaged, there's no easy way for translators to see the built-in Qt app translations until its packaged.
- According to [this](https://doc.qt.io/qt-5/internationalization.html#available-catalogs) there is `qt_` and `qtbase_` however `qt_zh_CN.qm` exists and `qtbase_zh_CN.qm` does not. Not sure what this means but it is at least better than nothing.

I force pushed an update to this which rebased it onto the current master and includes some changes I made to address these issues. The Qt translations are included at build time instead of requiring `windeployqt`. Maybe not ideal that the `qt_` file has to be listed explicitly...but think its a step in the right direction. In the end translations don't get added very routinely so its not that big of deal to list them explicitly.

inkydragon:
> `windeployqt`

Maybe: [probonopd/linuxdeployqt](https://github.com/probonopd/linuxdeployqt)


> no easy way for translators to see the built-in Qt app translations

Yes, but at the moment there is no good solution.


> however `qt_zh_CN.qm` exists and `qtbase_zh_CN.qm` does not. 

I've tried renaming `qt_zh_CN.qm` to `qtbase_zh_CN.qm` and it seems to work, at least for the print window.

dail8859:
The point I was making is that the changes I pushed should solve alot of the issues. It will not rely on `windeployqt` (or `linuxdeployqt`) for finding the correct translations since now it is just part of the build process to run `lrelease` on the Notepad Next `ts` files and also include the extra built-in Qt `.qm` files.

dail8859:
@inkydragon  Thanks for the PR!

PR #141, opened 2022-05-17
Add installation instructions to README

dail8859:


dail8859:
@zocker-160 Seem reasonable? Just wanted to keep it kind of short without going too in-depth.

zocker-160:
@dail8859 I think this is very reasonable, you both link to the Flathub page and provide the one line install which is very useful.

I think this is good.

PR #161, opened 2022-07-02
Include translations on Linux

### LINKED ISSUE #155: Application translation files are not included in AppImage
As discovered in #152 it looks like the AppImage has never included the Notepad Next translation files. Seems to be fine for Windows.

cc @eson57

dail8859:
Closes #155

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,6 +38,7 @@ jobs:
           rm -rf build-linux
           mkdir build-linux 
           cd build-linux
+          lrelease ../src/NotepadNext.pro
           qmake ../src/NotepadNext.pro
           make -j$(nproc)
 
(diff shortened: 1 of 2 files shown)


dail8859:
Might have to do something like this...

https://github.com/mapeditor/tiled/blob/a71fd94211e062b96d66d65b07cc384053bc70d9/src/tiled/languagemanager.cpp

elcuco:
Seems like work shifted to this PR: https://github.com/dail8859/NotepadNext/pull/318 - is it safe to close it?

dail8859:
Yeah I kept telling myself I'd pick this back up but never got around to it. Even if #318 wasn't open I'd still probably start again from scratch since I can't quite remember exactly what issues I was dealing with since this was only partially complete.

PR #180, opened 2022-07-26
Enable some NSIS Components by default

sitiom:
I think a start menu shortcut, context menu, and an auto-updater are what users would want by default in an installation. This is especially important for Winget users, as it runs silently by default, and they would expect it to appear in the Start Menu right after installation.


dail8859:
Awesome! Sounds good to me!

PR #182, opened 2022-07-26
Bring Window of primary instance to foreground on Linux

### LINKED ISSUE #149: Bring existing window to front primary instance is detected (ubuntu/gnome)
Running v0.5.1 on Ubuntu 18.04; when executing command to open file while existing instance is already running - it gets nicely added to tab list, but window is not brought to front (and one needs to manually dig it from below other applications).

```
[     0.216] I: =============================
[     0.216] I: NotepadNext v0.5.1
[     0.216] I: Build Date/Time: Apr 30 2022 19:14:55
[     0.216] I: Qt: 5.15.2
[     0.216] I: OS: Ubuntu 18.04.6 LTS
[     0.216] I: Locale: en_US
[     0.216] I: CPU: x86_64
[     0.216] I: File Path: /tmp/.mount_notepapY14eR/usr/bin/NotepadNext
[     0.216] I: Arguments: /tmp/.mount_notepapY14eR/AppRun.wrapped /tmp/a.txt
[     0.217] I: =============================
[     0.217] I: Primary instance already running. PID: 28998
[     0.220] I: Secondary instance closing...
```

dail8859:
Closes #149

### DIFF
--- a/src/NotepadNext/dialogs/MainWindow.cpp
+++ b/src/NotepadNext/dialogs/MainWindow.cpp
@@ -36,14 +36,10 @@
 
 #ifdef Q_OS_WIN
 #include <QSimpleUpdater.h>
-#endif
-
-#ifdef Q_OS_WIN
 #include <Windows.h>
 #endif
 
 #include "DockAreaWidget.h"
-#include "DockWidgetTab.h"
 
 #include "NotepadNextApplication.h"
 #include "Settings.h"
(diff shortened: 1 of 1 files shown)


src/NotepadNext/dialogs/MainWindow.cpp
```diff
@@ -1319,7 +1303,15 @@ void MainWindow::bringWindowToForeground()
 
         AttachThreadInput(GetWindowThreadProcessId(GetForegroundWindow(), nullptr), GetCurrentThreadId(), FALSE);
     }
+#else
+    show();
```
elcuco, line 1307: The linux code path seems to work for windows as well - tried it myself:

Scenario: I open NotepadNext, I open explorer - and while explorer is open, I drag a window into NotepadNext - with the "QtOnlyAPIs(tm)" - the window gets to the front. If I uncomment everything on this function - window still is on background.

Tested on Windows 10 with Qt 6.3.

Question: do you really need all that ifdef... or can you just keep the Linux branch on all platforms?

elcuco (commented):
... are you sure this is needed....?

src/NotepadNext/dialogs/MainWindow.cpp
dail8859: I can try it again on Qt6.3 when I get some free time but I believe the original need for this was in the case of using the right-click context menu to "Open with Notepad Next". Windows does not like it when an application has no interaction (i.e. not getting something dropped on it) and it wants to force its way to the foreground.

dail8859: Tried it again and there is 1 specific scenario that I just can't seem to handle with just Qt calls. If Notepad Next is in the background (e.g. another window is partially in front of it) and the user right-clicks a file and selects "Open with Notepad Next" then it won't change the z order of the application's window. Every other scenario I tested works, even if the application is minimized it will still restore itself and go to the foreground.

I even tried several suggestions here: https://stackoverflow.com/questions/6087887/bring-window-to-front-raise-show-activatewindow-don-t-work

So unfortunately I think this needs to keep the Win32 API calls.

elcuco (approved):
(not that my vote counts, but here, have an upvote!)

PR #235, opened 2022-10-17
Fix displaying of non-ascii chars in Hex Viewer

akmll:
In the Hex Viewer the char values bigger than 0x7f is interpreted as negative integer and is padded with leading 'FF' when displaying.

This commit cast the char value returned by scintilla to 'unsigned char' before formatting it to QString for displaying.
hexviewer


dail8859:
Awesome! Thanks for the PR!

Issue #77, opened 2022-04-08
AppImage is built on a too-new system


pgarciaq:
Running the AppImage on RHEL8 fails because the AppImage is built on a system with glibc 2.29, while RHEL8 contains glibc 2.28:

$ ./NotepadNext-x86_64.AppImage 
/tmp/.mount_NotepajreKDD/AppRun.wrapped: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by /tmp/.mount_NotepajreKDD/AppRun.wrapped)

$ rpm -qi glibc
Name        : glibc
Version     : 2.28
Release     : 164.el8_5.3
Architecture: x86_64

Please rebuild on an older Linux.

dail8859:
Thanks for opening this issue.

I made a quick attempt in #78 to see if just using an older Ubuntu version would work but no immediate luck.

I'd like to move to Qt 6.2 at some point in the near future. I'll have to invest further if there is some way to support Qt 6.2 with glibc 2.28

dail8859:
[Qt6.2 is linked against glibc 2.28](https://doc.qt.io/qt-6/supported-platforms.html#availability-of-packages)

> Note: Linux binary packages are linked against glibc 2.28. If you are using a glibc version older than that you need to build Qt from sources or upgrade glibc to 2.28 or newer.

And it looks like Ubuntu 18.04 uses glibc 2.27

Since Qt5 and Qt6 are currently supported for the time being, I might have to try building Qt5 on an older version of Linux and Qt6 on a newer version.

dail8859:
@pgarciaq 

Try downloading the Qt5 AppImage from this run (you'll probably have to unzip it first): https://github.com/dail8859/NotepadNext/actions/runs/2141001190

It is built with Ubuntu 18.04 and should be compatible

lxtwin:
QT5 works on RHEL 8.4, QT6 does not.

QT5_NotepadNext-x86_64.AppImage
[     0.037] I: Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
[     0.204] I: =============================
[     0.204] I: NotepadNext v
[     0.204] I: Qt: 5.15.2
[     0.204] I: OS: Red Hat Enterprise Linux 8.4 (Ootpa)
[     0.204] I: CPU: x86_64
[...]

PR #252, opened 2022-10-27
Switch to Qt6 for official distribution

dail8859:


.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         - 
           qt_version: "5.15.2"
           modules: ""
-          os: ubuntu-18.04
+          os: ubuntu-latest
```
elcuco, line 21: You can still use` os: ubuntu-20.04`. This seems to work for my branch and the generated AppImages work my Debian distro.

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
...
         - 
           qt_version: "6.2.4"
```
elcuco, line 23: `qt_version: "6.3.1"` - again, this seems to work for me.

.github/workflows/build.yml
```diff
@@ -200,21 +198,21 @@ jobs:
...
       - name: Upload Windows Installer
         run: |
-          gh release upload ${{ github.ref_name }} NotepadNext-Windows-Qt5.15.2-Installer/NotepadNext-${{ github.ref_name }}-Installer.exe
+          gh release upload ${{ github.ref_name }} NotepadNext-Windows-Qt6.2.4-Installer/NotepadNext-${{ github.ref_name }}-Installer.exe
 
       # The artifact is unzipped, so need to zip it back up so it can be uploaded
       - name: Upload Windows Zip
         run: |
-          cd NotepadNext-Windows-Qt5.15.2-Zip/
+          cd NotepadNext-Windows-Qt6.2.4-Zip/
```
elcuco, line 206: Just call it `Qt6` - this way the next time you upgrade Qt, less to modify.

elcuco (commented):
IMHO you can bump versions even nigher.
I don't see OSX builds here, am I missing something? How about windows?

.github/workflows/build.yml
dail8859: I switched back to 18.04 due to #77 needing an older version of GCC for Qt5. Unless there is a reason for 20.04 then I'm fine using `ubuntu-latest`.

.github/workflows/build.yml
dail8859: I plan on adding 6.3.1 as well soon. My plan is to at least build on one LTS version (6.2.4 in this case) along with the latest available Qt version. I'll keep 5.15 for now due to its extended LTS, but if it ever starts causing pains to develop and support it (work arounds or lots of `#ifdefs` then it will be dropped).

.github/workflows/build.yml
dail8859: The output directories are based on the `matrix.config.qt_version` variable so it has the full version number in there. Plus if there is a 6.2.4 and 6.3.1 then they'd have to be different. Something like string splitting to have `Qt6.2` and `Qt6.3` might be a viable option though.

dail8859:
The GH action shows it is building for mac fine.

.github/workflows/build.yml
elcuco: Are you sure binaries made on Ubuntu 22.04 will run on 20.04? Are you requiring a newer ubuntu?

dail8859: > Are you sure binaries made on Ubuntu 22.04 will run on 20.04?

The short answer is: no.

That's the main reason for this draft PR so that if there are cases like this that I'm unaware of then we can make sure that this is accounted for. If there is a known justification to not use the latest ubuntu version then I'm perfectly fine with that.

PR #257, opened 2022-11-03
Redeuce some warnings

elcuco:
Some unused functions, and arguments.
As I am using a C++17 switch, I am a little worried...

Is C++17 the version we are aiming for...?

### DIFF
--- a/src/NotepadNext/ScintillaNext.h
+++ b/src/NotepadNext/ScintillaNext.h
@@ -27,10 +27,6 @@
 #include <QFileInfo>
 
 
-static bool isRangeValid(const Sci_CharacterRange &range)
-{
-    return range.cpMin != INVALID_POSITION && range.cpMax != INVALID_POSITION;
-}
 
 
 class ScintillaNext : public ScintillaEdit
@@ -38,6 +34,11 @@ class ScintillaNext : public ScintillaEdit
     Q_OBJECT
 
 public:
+    static bool isRangeValid(const Sci_CharacterRange &range)
+    {
+        return range.cpMin != INVALID_POSITION && range.cpMax != INVALID_POSITION;
+    }
+
     explicit ScintillaNext(QString name, QWidget *parent = Q_NULLPTR);
     virtual ~ScintillaNext();
 
--- a/src/NotepadNext/docks/LuaConsoleDock.cpp
+++ b/src/NotepadNext/docks/LuaConsoleDock.cpp
@@ -191,6 +191,7 @@ LuaConsoleDock::LuaConsoleDock(LuaState *l, QWidget *parent) :
     input->installEventFilter(this);
 
     connect(input, &ScintillaNext::updateUi, [=](Scintilla::Update flags) {
+        Q_UNUSED(flags);
         int curPos = input->currentPos();
         int bracePos = INVALID_POSITION;
 
--- a/src/NotepadNext/LuaExtension.cpp
+++ b/src/NotepadNext/LuaExtension.cpp
@@ -44,7 +44,7 @@ IFaceTableMixer ifacemixer;
 static bool call_function(lua_State *L, int nargs, bool ignoreFunctionReturnValue);
 
 // Helper function from SciTE
-static int Substitute(std::string &s, const std::string &sFind, const std::string &sReplace) {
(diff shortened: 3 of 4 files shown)


dail8859:
> As I am using a C++17 switch, I am a little worried...

Why is that exactly?

> Is C++17 the version we are aiming for...?

That's currently what is being used for Windows/MSVC. If there's a need or reason to switch to a different version I'm willing to consider that. I'm not familiar with what other OS build chains support easily.

src/NotepadNext/ScintillaNext.h
```diff
@@ -27,7 +27,7 @@
 #include <QFileInfo>
 
 
-static bool isRangeValid(const Sci_CharacterRange &range)
+[[maybe_unused]] static bool isRangeValid(const Sci_CharacterRange &range)
```
dail8859, line 30: Would keeping the definition in the header and then moving the implementation to the cpp file fix this warning? Because it is definitely used.

src/NotepadNext/docks/LuaConsoleDock.cpp
```diff
@@ -215,6 +215,8 @@ LuaConsoleDock::LuaConsoleDock(LuaState *l, QWidget *parent) :
         else {
             input->braceHighlight(INVALID_POSITION, INVALID_POSITION);
         }
+
+        Q_UNUSED(flags);
```
dail8859, line 219: Move this to the top of the lambda. Thats just kind of the way I've been doing it in other places so that will just help keep things consistent.

dail8859 (changes requested):
Honestly if the code is actually unused, just go ahead and delete the function/method. Especially the LuaExtension stuff...that part has always been a bit messy but not really looking to put alot of effort into it incase it is not a long-term solution, so removing it is fine.

src/NotepadNext/ScintillaNext.h
elcuco: but then it would not be inline. Do you need this to be inline?

dail8859: No. There's no known reason why this would be required to be inlined.

elcuco: This is a very small function, inlining it seems reasonable (creating a stack, and a call for such a trivial function seems "lame"). I have another approach. See the new commit.

src/NotepadNext/docks/LuaConsoleDock.cpp
elcuco: Removing the flag/variable name will have the same effect (I am in new phase, in which I want to reduce Qt macros). See new commit.

src/NotepadNext/ScintillaNext.h
dail8859: The static method is fine with me. Honestly, I wouldn't be surprised if the compiler optimizes this into an inline call anyways.

src/NotepadNext/docks/LuaConsoleDock.cpp
dail8859: Personally I've always used the `Q_UNUSED` macro rather than removing the variable name (technically you can also remove the entire parameter, but I don't like that idea).

I'm not against removing the variable name, but for sake of consistency let's keep the macro. If for some reason later its decided to forego the macro route, then some other option like you suggested can be applied across the entire code base.

Even if something is a little ugly, I'd rather lean towards the side of being explicit.

src/NotepadNext/ScintillaNext.h
elcuco: This is why I kept it on the header, if it was on another compilation unit, the compiler will not be able to optimize it :)

src/NotepadNext/docks/LuaConsoleDock.cpp
elcuco: done. building and in 10 minutes we should get a green light.

dail8859: > but for sake of consistency let's keep the macro

I'm not seeing this in the latest changes.

elcuco: arg.. internet issues (rebased + squashed now): https://github.com/dail8859/NotepadNext/pull/257/files#diff-806bf892a96cd6a0af53c1c14f08bd10b2733e4cd12e71c26b998509a3f7971dR194

Issue #171, opened 2022-07-09
Dark mode


Jogai:
My system is using mostly QT apps and its configured with a dark theme that every qt app respects, even xnviewmp with system libs, which doesnt have a dark mode in windows i believe. However notepadNext doesnt. I dont know if its something that interests you, but I believe having a dark mode will give NotepadNext a popularity boost.

FollowerOfBigboss:
Hi @Jogai, i think you should extend issue as "theme support". It shouldn't hard to implement dark mode support if you have some qt knowledge. For a simple solution you may invoke setStyleSheet method.

https://doc.qt.io/qt-6/stylesheet-examples.html

I would like to come with proof of concept pr but unfortunately my computer is broken. I believe you can implement a simple dark theme by trial and error before @dail8859 step up. Good luck.

dail8859:
My understanding is that using the Flatpak will follow the system Qt theme. The AppImage will not.

That being said...having user defined themes (eg loading a css file) would be good to have eventually.

Jogai:
I didnt use flatpak or appimage, I took it from the aur, which is a source build: https://aur.archlinux.org/packages/notepadnext

dail8859:
I have no clue how that works since it's not an official distribution mechanism for this project.

dail8859:
You can technically already launch the application with a custom CSS stylesheet by [command line](https://doc.qt.io/qt-5/qapplication.html#QApplication).

`NotepadNext.exe -stylesheet style.css`

You just need to create a `style.css` file in the same directory as the application. You can of course name the file something different or put it in a different location and specify the path on the command line.

It won't be perfect because the application uses its own stylesheet internally for tweaking some elements. But you can do some things quickly to test with, for example:

![image](https://user-images.githubusercontent.com/3694843/179326841-e77b9151-f722-4967-8d6f-6ff5291524a2.png)

Jogai:
Well, that's an option, but I was thinking more along the lines of this: https://successfulsoftware.net/2021/03/31/how-to-add-a-dark-theme-to-your-qt-application/

dail8859:
@Jogai

That points out a few ways to auto detect the OS preference but really just boils down to loading the appropriate CSS file.

The example I showed above was obviously very ugly and not suggesting using a red theme.

dail8859:
@Jogai Awesome! Glad to hear it is working now.

sonichy:
[...]

Issue #109, opened 2022-04-19
Tab filename poorly integrates with Breeze Dark


KrazyKirby99999:
Having installed the linux flatpak version, I found that the tabs containing the filename of the open file are difficult to read.

![Screenshot_20220419_163356](https://user-images.githubusercontent.com/69373044/164106262-dc57e142-ac54-49fb-82f9-666a7d0a99c2.png)

My desktop environment is KDE Plasma 5.24.4 and I am using the built-in Breeze Dark theme.

Satharus:
I have the same issue here. I am using the [Nordic Theme](https://store.kde.org/s/KDE%20Store/p/1327093/).

![image](https://user-images.githubusercontent.com/29730368/164114435-be260bd7-6f21-478d-b304-fde526c1e586.png)

dail8859:
Just some additional info...the tabbing/docking functionally comes from the [Advanced Docking System](https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System) (ADS), and I also apply a custom [stylesheet](https://github.com/dail8859/NotepadNext/blob/91949c7d6a62734713d44f35394031d05bbbf710/src/NotepadNext/stylesheets/npp.css) to it. So I'm not sure if this is an ADS issue, or styling issue currently.

mutageneral:
I have the same issue here. I am on the GNOME desktop environment and using the Dark Adiwata GTK theme ported to QT for a seamless experience between GTK and QT applications.

The application when in focus:
![image](https://user-images.githubusercontent.com/62779748/164132269-e199f36e-6f28-4b82-8375-36c324d42639.png)

The application when not in focus. Notice that now the text label is now readable: 
![image](https://user-images.githubusercontent.com/62779748/164132771-11308e6c-0915-4397-ab17-cc8d59af66d5.png)

QT theme settings:
![gnome-shell-screenshot-cg1xsm](https://user-images.githubusercontent.com/62779748/164132122-3109b0c1-6089-4e5b-a04f-175c26b9839e.png)

mutageneral:
Just downloaded the AppImage to see if it has the same issue, but it seems that the AppImage ignores my system theme and uses a built-in one. Is this the desired behaviour? Usually it is Flatpaks that have a hard time using system themes, not AppImages...
Tell me if you want me to open an issue for this specifically.
[...]

PR #264, opened 2022-11-10
Restore opened files between application restarts

dail8859:
This is a first attempt at saving/loading a user's session between application restarts.

This feature is **not** enabled by default. This needs enabled through the preferences dialog.

~~This feature does **not** currently save temporary files (that will come later). This base functionality is needed before temporary files can be properly saved/loaded.~~

TODO:
- [x] Make sure new files don't get named the same as existing files. E.g. since `New 1` can be loaded from the session, `File > New` would then also create `New 1`
- [x] Allow to files with the same name to be stored in the session. `C:\dir1\test.txt` and `C:\dir2\test.txt` both get saved as `text.txt` and will overwite each other.
- [x] Remove commented out code

### DIFF
--- a/src/NotepadNext/EditorManager.h
+++ b/src/NotepadNext/EditorManager.h
@@ -35,7 +35,6 @@ public:
 
     ScintillaNext *createEditor(const QString &name);
     ScintillaNext *createEditorFromFile(const QString &filePath, bool tryToCreate=false);
-    //ScintillaNext *cloneEditor(ScintillaNext *editor);
 
     ScintillaNext *getEditorByFilePath(const QString &filePath);
 
--- a/src/NotepadNext/EditorManager.h
+++ b/src/NotepadNext/EditorManager.h
@@ -33,18 +33,19 @@ class EditorManager : public QObject
 public:
     explicit EditorManager(QObject *parent = nullptr);
 
-    ScintillaNext *createEmptyEditor(const QString &name);
+    ScintillaNext *createEditor(const QString &name);
     ScintillaNext *createEditorFromFile(const QString &filePath, bool tryToCreate=false);
-    ScintillaNext *cloneEditor(ScintillaNext *editor);
+    //ScintillaNext *cloneEditor(ScintillaNext *editor);
 
     ScintillaNext *getEditorByFilePath(const QString &filePath);
+
+    void manageEditor(ScintillaNext *editor);
 
 signals:
     void editorCreated(ScintillaNext *editor);
     void editorClosed(ScintillaNext *editor);
 
 private:
-    void manageEditor(ScintillaNext *editor);
     void setupEditor(ScintillaNext *editor);
     void purgeOldEditorPointers();
 
--- a/src/NotepadNext/SessionManager.h
+++ b/src/NotepadNext/SessionManager.h
@@ -1,0 +1,38 @@
+/*
+ * This file is part of Notepad Next.
+ * Copyright 2022 Justin Dailey
+ *
+ * Notepad Next is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Notepad Next is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Notepad Next.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+
+#ifndef SESSIONMANAGER_H
+#define SESSIONMANAGER_H
+
+#include <QVector>
+
+
+class ScintillaNext;
+class EditorManager;
+
+class SessionManager
+{
+public:
+    SessionManager();
+
+    static void SaveSession(QVector<ScintillaNext *> editors);
+    static void LoadSession(EditorManager *editorManager);
+};
+
+#endif // SESSIONMANAGER_H
--- a/src/NotepadNext/SessionManager.h
+++ b/src/NotepadNext/SessionManager.h
@@ -55,6 +55,8 @@ public:
 private:
     QDir sessionDirectory() const;
 
+    void saveIntoSessionDirectory(ScintillaNext *editor, const QString &sessionFileName) const;
+
     SessionFileType determineType(ScintillaNext *editor) const;
 
     void clearSettings() const;
--- a/src/NotepadNext/SessionManager.h
+++ b/src/NotepadNext/SessionManager.h
@@ -20,7 +20,9 @@
 #ifndef SESSIONMANAGER_H
 #define SESSIONMANAGER_H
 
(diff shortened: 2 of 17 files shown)


src/NotepadNext/SessionManager.h
```diff
@@ -0,0 +1,41 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Notepad Next.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+
+#ifndef SESSIONMANAGER_H
```
elcuco, line 20: You should start using `#pragma once` it does the same, and is supported on GCC since forever. Do you want me to handle this in another PR?

dail8859: > is supported on GCC since forever

Interesting. Guess I always thought of `#pragma once` as a MSVC++ specific feature (I first found it many years ago!), but as you mentioned it appears it has been universally supported for a while now.

So yes I agree `#pragma once` is desirable. You are certainly welcome to submit the changes. Only need to worry about code under `.src/NotepadNext`

mikelupe:
Very nice, will be my daily driver - thanks for your efforts.

One situation ("bug"): 
0. Fresh start, no sessions apart "New 1"
1. "New 1" > type some text > "Save as.." test.txt
2. Double click to create a new tab > "New 1" > type some text > close NotepadNext w/o saving
3. Relaunch NotepadNext ; State: one tab "test.txt" (red with content), one tab "New 1" (red with content)
4. Again double click to create a new tab > "New 1" (blue empty)
5. On close and reopen, only the last "New 1" is kept, the original "New 1" (red) is ~overwritten

Basically: As long as a "New X" tab hasn't been saved, the next increment of new still is "New X", so no increment :)
I hope it's reproducable

kind regards

dail8859:
@mikelupe Thanks for testing it out!

> 1. "New 1" > type some text > "Save as.." test.txt
> 2. Double click to create a new tab > "New 1" > type some text > close NotepadNext w/o saving

Just to make sure, in step 2 wouldn't it be "New 2" since the "New 1" document was already saved?

---

But yes this is one of the bugs, that you can't have two files with the same name currently. A couple points to this:

- If `New 2` (or any other temp file for that matter) is loaded, then a new tab (e.g. File > New) should never be named that, so some logic would need added to keep incrementing the number in the name until one is found that doesn't exist.
- The same thing will happen if you have `C:\dir1\test.txt` and `C:\dir2\test.txt` since they both tried to get saved in the session folder both as `test.txt`. I've got an idea to fix this.

mikelupe:
> Just to make sure, in step 2 wouldn't it be "New 2" since the "New 1" document was already saved?

There would be no more "New 1" as it had been saved as "test.txt" in step 1 

Thanks again for your effort, this will be epic :)

mikelupe:
I checked your latest commit:

I had to remove the rm ~/.config/NotepadNext/NotepadNext.ini , otherwise I got a segfault :) 

Unfortunately now it generates a "New 2" but then later on again another "New 2". Almost there

dail8859:
> otherwise I got a segfault :)

Yeah that makes sense, but would indicate some better error handling is needed incase the settings file gets modified or corrupted.

> Unfortunately now it generates a "New 2" but then later on again another "New 2". Almost there

Yep the root issue was fixed so that even if you have two `New 2` temp files they won't overwrite each other. Still need to fix the file name incrementing to prevent this.

Just curious since you are on a non-Windows machine, where exactly does it save the session files too? I'm assuming something like `~/.config/NotepadNext/NotepadNext/session/`? I've only tested it on Windows.

mikelupe:
I see, nice.

Re location - I had to find it first :) It saves it to "~/.local/share/NotepadNext/NotepadNext/session/" . so $HOME/.local/...

dail8859:
> It saves it to "~/.local/share/NotepadNext/NotepadNext/session/" . so $HOME/.local/...

Hmmm, ok. The Qt documentation for standard paths is a bit fuzzy, so it saved to a semi-sane location on Windows but didn't know what that would look like on other systems.

mikelupe:
I think that's quite legit. ~/.config/NotepadNext/session/ would be ok as well. Eventually a config setting with the recent defaults could make sense.

mikelupe:
Nice

src/NotepadNext/EditorManager.h
```diff
@@ -33,18 +33,19 @@ class EditorManager : public QObject
 public:
     explicit EditorManager(QObject *parent = nullptr);
 
-    ScintillaNext *createEmptyEditor(const QString &name);
+    ScintillaNext *createEditor(const QString &name);
     ScintillaNext *createEditorFromFile(const QString &filePath, bool tryToCreate=false);
-    ScintillaNext *cloneEditor(ScintillaNext *editor);
+    //ScintillaNext *cloneEditor(ScintillaNext *editor);
```
elcuco, line 38: you "should" clean this up

dail8859: Agreed

dail8859: Fixed

mikelupe:
Thanks for your speedy work, and for the mention :) The latest build is becoming my driver. 

There is one thing though, double clicking generates incremented "New X", but if closed w/o saving, on a new creation it still increments - so create "New 2" > close "New 2", create a new one gives "New 3" > close w/o saving > create a new one gives "New 4" etc. 

Would it be easy to check on this, even if it's not really a show-stopper? It would make it even more consistent.

dail8859:
> There is one thing though, double clicking generates incremented "New X", but if closed w/o saving, on a new creation it still increments - so create "New 2" > close "New 2", create a new one gives "New 3" > close w/o saving > create a new one gives "New 4" etc.

This is actually a preexisting bug (not sure if this PR made it worse or not) but even in the official v0.5.6 release I can recreate it by:

1. Start a fresh instance, which just contains `New 1`
2. `File > New` creates `New 2`
3. `File > Close All` closes files and creates `New 3`
4. Repeat step 3, creates `New 4` etc.

This was one of the bugs I ran across and thought I introduced it with these changes but there is at least some not-desirable behavior that already exists. Appreciate you testing things and providing feedback!

mikelupe:
Let's be frank, It's not really a bug, maybe just some polishing.

Something else, would a dark mode be a lot of work? But this would extend it over original Notepad++'s features.

Thanks again for your work

dail8859:
> Let's be frank, It's not really a bug, maybe just some polishing.

Agreed

> Something else, would a dark mode be a lot of work?

Yes and no. Getting something basic is easy, to really polish it would probably take a bit of time. There may be existing 3rd party dark theme css files that would work well but I haven't tested them. You can load custom css files: see https://github.com/dail8859/NotepadNext/issues/171#issuecomment-1186036612

Not something I'll be working on myself any time soon.

There are other non-Windows systems that have Qt theming built-in so it carries over to Notepad Next: https://github.com/dail8859/NotepadNext/issues/109#issuecomment-1105907443

mikelupe:
Thanks for the hint.

As it's mainly about the editor background that would need to be dark, I'll be waiting. There really is no hurry.

Issue #216, opened 2022-09-30
optimizing the find and replace text box


tangkeplus:
in Mac
The find and replace text field is too short
If the text is longer, there is too little visible text

dail8859:
I do not have a Mac.

Can you post a screenshot of the issue?

PaddyTB:
Here you go, from my Mac:  Paddy From: tangkeplus ***@***.***>Date: Friday, 30 September 2022 at 09:27To: dail8859/NotepadNext ***@***.***>Cc: Subscribed ***@***.***>Subject: [dail8859/NotepadNext] optimizing the find and replace text box (Issue #216)in MacThe find and replace text field is too shortIf the text is longer, there is too little visible text—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>

elcuco:
Can you check if this build fixes the issue for you?
https://github.com/dail8859/NotepadNext/suites/9491423888/artifacts/450040615

elcuco:
ping... can you test this version and see if the problem is fixed?

dail8859:
@elcuco

Since nobody can verify the fix, what are your thoughts? If you feel it will either fix this issue or at least won't have a negative impact, then I'm fine merging it. If you are bit unsure then maybe we are better off closing this issue along with the PR. I'm fine either way.

elcuco:
I asked a friend with a Mac to test the latest artifact from the git, (https://github.com/dail8859/NotepadNext/suites/10093757647/artifacts/491972251), he claims it works fine for him. IMHO, the Qt6 version fixed the issue. Close it.

@tangkeplus please re-open this if after the next update you still see it.

dail8859:
@elcuco Thanks!

Issue #136, opened 2022-05-10
Incremental filtering support


data-man:
SciTE has this feature:

![ksnip_20220510-083755](https://user-images.githubusercontent.com/1754269/167537912-58d51ad6-e79b-47b9-9185-fe1f082a1d31.png)

![ksnip_20220510-084123](https://user-images.githubusercontent.com/1754269/167538200-7bdb24c2-80e1-40a2-8331-9e9bc90deab9.png)

The nice project, thank you!

dail8859:
I'm not against this feature but I'm converting this over to a discussion for anyone interested in it.

Issue #132, opened 2022-05-06
Find/Replace Dialog layout broken on Linux


pienaarp:
Running Notepad Next on Centos 8 Stream - the find/replace dialog is a bit messed up.

<img width="617" alt="Find_Dialog" src="https://user-images.githubusercontent.com/8780803/167105500-ce5ebe2b-4593-4d41-94bd-3e964278b2ea.png">

dail8859:
@pienaarp Are you using the AppImage or Flatpak?

pienaarp:
AppImage

dail8859:
Hmmm...that's odd. Not familiar with CentOS. Is there some kind of high DPI or scaling you can adjust to see if that affects it?

What happens if you resize the dialog larger? Does that make it look better?

sonulohani:
Not able to reproduce in ubuntu

vfedyay:
Checked this on Debian11 with KDE
screen resolution 1920x1200
I use AppImage

Problem is on Replace tab only.
![image](https://user-images.githubusercontent.com/11298337/167264948-b877284f-18f2-4350-9ea4-d21ecd3517b8.png)
![image](https://user-images.githubusercontent.com/11298337/167264956-caf79afe-7014-4039-a960-71474c3d410c.png)

Tried on Debian11 with XFCE - works perfectly
![image](https://user-images.githubusercontent.com/11298337/167264985-f3c43f13-2da8-4ff2-a9b5-97e0c95c431d.png)

And this window can not be resized vertically but horizontally only.

pienaarp:
FWIW I am running MATE desktop on Centos 8 Stream.

"What happens if you resize the dialog larger? Does that make it look better?" <- As Volodymyr mentioned, can resize horizontally but not vertically.

dail8859:
I'm assuming this is because I'm forcing a specific height for the dialog and the applied theme doesn't work for the limited space.

I'll look into some way to dynamically calculate a default height, but I have a hard time testing this since I can't reproduce this on Windows easily.

pienaarp:
If you are supporting Linux (for which I am grateful) might be worth having
a couple of virtual machines for testing on Linux desktops ?

dail8859:
@pienaarp 

> If you are supporting Linux (for which I am grateful) might be worth having
a couple of virtual machines for testing on Linux desktops ? 

[...]

PR #280, opened 2022-11-24
Attempt to fix find/replace dialog

elcuco:
It seems that under some scenarios some edit boxes are too small. Lets see if this works for everyeone (tested on Windows Qt6 and looks good).

Will close https://github.com/dail8859/NotepadNext/issues/216

### DIFF
--- a/src/NotepadNext/dialogs/FindReplaceDialog.ui
+++ b/src/NotepadNext/dialogs/FindReplaceDialog.ui
@@ -355,6 +355,9 @@
        </property>
        <item>
         <layout class="QFormLayout" name="formLayout">
+         <property name="sizeConstraint">
+          <enum>QLayout::SetMinimumSize</enum>
+         </property>
          <property name="horizontalSpacing">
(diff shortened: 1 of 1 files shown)


dail8859:
Thanks for the PR. Haven't had a chance to test it out yet but seems simple enough. Would this potentially fix #136 as well?

dail8859:
Sizing policies are still a bit of a mystery to me. I tried out the PR on Windows and don't see any negative effects. As soon as someone can confirm that this fixes #216 or #132 (better yet both) then I'll merge this.

dail8859 (approved):

elcuco:
> Thanks for the PR. Haven't had a chance to test it out yet but seems simple enough. Would this potentially fix #132 as well?

I don't think so. The problem in #132 is vertical space - which means the window is not properly sized.  This might be a Qt5/Qt6 issue tough.

elcuco:
lets close it, as IMHO the problems will get fixed by qt6.

Issue #305, opened 2023-01-19
No %f in the .desktop file Exec


pranatahouse:
I installed the AppImage version using AppImageLauncher to automate creating the desktop file on Ubuntu. The problem is that when i choose open with other application to a c++ file for example, NotepadNext is not listed there because it does not have %f to it. So for now i edit the desktop file like this to make it visible on open with dialog.

Exec="/appimage/location/NotepadNext-x86_64_08f43ec222d8caa1af2101d59f80a34f.AppImage" %f

elcuco:
The desktop file used inside an app image, (IMHO) is auto-generated from AppImage builder:

https://github.com/dail8859/NotepadNext/blob/4ddddcbf31ab8d1d37f0719a0242d87496c9d4dc/.github/workflows/build.yml#L55-L65

We need to identify what generates that desktop file? It is our desktop file? Is such case, the fix is here:
https://github.com/dail8859/NotepadNext/blob/4ddddcbf31ab8d1d37f0719a0242d87496c9d4dc/deploy/linux/NotepadNext.desktop#L2

dail8859:
@elcuco Thanks for the PR!

> We need to identify what generates that desktop file? It is our desktop file?

That was my thought as well. I took a look at a few different repos and they all have the `%f` in the exec command in the desktop file. So in the end I think the PR is appropriate, but would still be interested if it fixes the issue with the AppImage.

--- 

@pranatahouse Try this AppImage from the PR @elcuco created: https://github.com/dail8859/NotepadNext/suites/10529939256/artifacts/523291809

pranatahouse:
> @elcuco Thanks for the PR!
> 
> > We need to identify what generates that desktop file? It is our desktop file?
> 
> That was my thought as well. I took a look at a few different repos and they all have the `%f` in the exec command in the desktop file. So in the end I think the PR is appropriate, but would still be interested if it fixes the issue with the AppImage.
> 
> @pranatahouse Try this AppImage from the PR @elcuco created: https://github.com/dail8859/NotepadNext/suites/10529939256/artifacts/523291809

ok i'll try it out.

pranatahouse:
Thank you @dail8859 i've tested the appimage and now it's working fine installed using AppImageLauncher.

dail8859:
@pranatahouse Thanks for testing it out!

@elcuco Thanks for the PR. Will get that merged now.

PR #306, opened 2023-01-23
Fix for #305 - pass arguments to executable, from desktop file

elcuco:
(title is all you need).


dail8859 (approved):
LGTM. I'll plan to merge this in either way, but will wait to see if it fixes #305.

PR #316, opened 2023-02-01
Update i18n *.ts files

inkydragon:
just run `lupdate`

Translation updates in another pr.


dail8859:
@inkydragon Thanks for the PR!

I tend to forget about the translation file updates. Might be worth me looking into a manual or cronjob type of GitHub action to run `lupdate` and push these changes back to the repository so things stay more up to date.

PR #318, opened 2023-02-01
Add support load translation for linux

wineee:
Closes https://github.com/dail8859/NotepadNext/issues/155

Translation file should be installed in `XDG_DATA_DIRS`/`APPNAME`/translations， `XDG_DATA_DIRS` Should default to /usr/local/share:/usr/share.

QStandardPaths::AppDataLocation will read `xdgDataDirs/organizationName/applicationName`

https://github.com/qt/qtbase/blob/1c6bf3e09ea9722717caedcfcceaaf3d607615cf/src/corelib/io/qstandardpaths_unix.cpp#L385-L390
https://github.com/qt/qtbase/blob/1c6bf3e09ea9722717caedcfcceaaf3d607615cf/src/corelib/io/qstandardpaths_unix.cpp#L27-L38

### DIFF
--- a/src/NotepadNext/NotepadNextApplication.cpp
+++ b/src/NotepadNext/NotepadNextApplication.cpp
@@ -36,6 +36,8 @@
 
 #include <QCommandLineParser>
 #include <QSettings>
+#include <QStandardPaths>
+#include <QLibraryInfo>
 
 #ifdef Q_OS_WIN
 #include <Windows.h>
@@ -390,11 +392,22 @@ void NotepadNextApplication::loadTranslation(QLocale locale)
     qInfo(Q_FUNC_INFO);
 
     // TODO: look into QLibraryInfo::location(QLibraryInfo::TranslationsPath)?
+#ifdef Q_OS_LINUX
+    QString languagePath = QStandardPaths::locate(QStandardPaths::AppDataLocation, QString("translations"), QStandardPaths::LocateDirectory);
+#else
     const QString languagePath = QApplication::applicationDirPath() + "/i18n";
+#endif // Q_OS_LINUX
 
     // Load translation for NotepadNext
     //  e.g. `i18n/NotepadNext.en.qm`
-    if (translatorNpn.load(locale, QApplication::applicationName(), QString("."), languagePath)) {
+    bool isTransloaded = translatorNpn.load(locale, QApplication::applicationName(), QString("."), languagePath);
+#ifdef Q_OS_LINUX
+    if (!isTransloaded) {
+      languagePath = QApplication::applicationDirPath() + "/../share/NotepadNext/NotepadNext/translations";
+      isTransloaded = translatorNpn.load(locale, QApplication::applicationName(), QString("."), languagePath);
+    }
+#endif // Q_OS_LINUX
+    if (isTransloaded) {
         installTranslator(&translatorNpn);
         qInfo("Loaded %s translation for Notepad Next", qUtf8Printable(locale.name()));
     } else {
@@ -403,6 +416,9 @@ void NotepadNextApplication::loadTranslation(QLocale locale)
 
     // Load translation for Qt components
     //  e.g. `i18n/qt_en.qm`
+#ifdef Q_OS_LINUX
+    languagePath = QLibraryInfo::location(QLibraryInfo::TranslationsPath);
+#endif // Q_OS_LINUX
     if (translatorQt.load(locale, QString("qt"), QString("_"), languagePath)) {
         installTranslator(&translatorQt);
         qInfo("Loaded %s translation for Qt components", qUtf8Printable(locale.name()));
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,9 @@ jobs:
           cd build-linux
           make install INSTALL_ROOT=AppDir
           cd NotepadNext
+          lrelease ../../src/NotepadNext.pro
+          mkdir -p AppDir/usr/share/NotepadNext/NotepadNext/translations
+          cp ../../i18n/NotepadNext*.qm AppDir/usr/share/NotepadNext/NotepadNext/translations/
           wget --no-verbose "https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage"
           wget --no-verbose "https://github.com/linuxdeploy/linuxdeploy-plugin-qt/releases/download/continuous/linuxdeploy-plugin-qt-x86_64.AppImage"
           chmod +x linuxdeploy*.AppImage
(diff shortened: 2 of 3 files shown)


src/NotepadNext/NotepadNextApplication.cpp
```diff
@@ -403,6 +409,9 @@ void NotepadNextApplication::loadTranslation(QLocale locale)
 
     // Load translation for Qt components
     //  e.g. `i18n/qt_en.qm`
+#ifdef Q_OS_LINUX
+    languagePath = QLibraryInfo::location(QLibraryInfo::TranslationsPath);
```
wineee, line 413: QLibraryInfo::TranslationsPath should work on Windows and MacOS, but I didn't test it. Currently only used in linux.

dail8859:
@wineee Awesome thanks for the PR! My understanding is that this will fix #155 which is specific to the AppImage, correct?

src/NotepadNext/NotepadNextApplication.cpp
dail8859: No problem. When I get some free time I'll test it out on Windows unless anyone else gets to it before I do :)

wineee:
I test in NixOS,  haven't tested AppImage yet. The path of AppImage is special, not sure if it works, I will try it tomorrow

dail8859:
My understanding was that for Linux, loading translations worked on the flatpak distribution but didn't work with the AppImages. Wasn't aware of any other Linux translation issues.

src/NotepadNext/NotepadNextApplication.cpp
```diff
@@ -390,11 +392,22 @@ void NotepadNextApplication::loadTranslation(QLocale locale)
...
     const QString languagePath = QApplication::applicationDirPath() + "/i18n";
+#endif // Q_OS_LINUX
 
     // Load translation for NotepadNext
     //  e.g. `i18n/NotepadNext.en.qm`
-    if (translatorNpn.load(locale, QApplication::applicationName(), QString("."), languagePath)) {
+    bool isTransloaded = translatorNpn.load(locale, QApplication::applicationName(), QString("."), languagePath);
+#ifdef Q_OS_LINUX
+    if (!isTransloaded) {
+      languagePath = QApplication::applicationDirPath() + "/../share/NotepadNext/NotepadNext/translations";
```
wineee, line 406: This work for me
```
[     0.038] D: TEST PATH:  "/home/rewine/.cache/appimage-run/b4672722b5b72a97c2b9caa2f923cb8620f21e365648aa68c96b865965e8bc25/usr/bin/../share/NotepadNext/NotepadNext/translations"
[     0.038] I: Loaded zh_CN translation for Notepad Next
[     0.038] D: QT TEST PATH:  "/home/rewine/.cache/appimage-run/b4672722b5b72a97c2b9caa2f923cb8620f21e365648aa68c96b865965e8bc25/usr/bin/../translations" # note：only find qtbase_xxx.qm in this directory
[     0.038] I: zh_CN translation not found for Qt components
[     0.038] I: LuaState::LuaState()
[     0.045] I: MacroManager::MacroManager(QObject*)

```

wineee:
This patch should work on all FHS(Filesystem Hierarchy Standard) compliant distributions, for example ：deb (debian/ubuntu/deepin), rpm(fedora/opensuse), pkgbuild(Arch Linux)

nix/flatpak/snap/AppImages are special packaging systems，  can't use strict FHS path. 

This patch also work if [XDG_DATA_DIRS](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) environment variables are set correctly,

src/NotepadNext/NotepadNextApplication.cpp
wineee: 图片

dail8859:
I spent some time digging into this. I think I'm just as confused as when I started. As far as Windows is concerned it seems straight forward enough (I'm assuming MacOS works the same way).

Linux of course is the difficult issue. AppImage, Flatpak, nixOS, XDG_DATA_DIRS, etc can all make things quite difficult to ensure things are working properly to load translation files.

I'm not sure if embedding the translation files is the right/quick solution. I've been trying to find other mature Qt projects to see how they handle it but nothing seems to be obvious.

wineee:
There seems to be no standard way:

Some applications find  translation files  from `XDG_DATA_DIRS`/`APPNAME`/translations: https://github.com/linuxdeepin/dtkwidget/blob/92a7561e0f541e165e91eb108cb846f0d4701fe3/src/widgets/dapplication.cpp#L757

Some applications use relative paths： https://github.com/pbek/QOwnNotes/blob/230c2b40dbe0f32edb3fbcd1a96dbd5ec6abcb24/src/main.cpp#L46-L54

Some applications embed translation files：https://github.com/cpeditor/cpeditor/blob/6722c603a322a581b5886a9ea4a2a7e67c2cb786/src/Core/Translator.cpp#L63

Some applications  just use hardcoded path（this cant work for AppImages）
https://github.com/linuxdeepin/dde-appearance/blob/5dbdee7c8c4da8817a1000360f50c6598a158719/src/service/main.cpp#L12

wineee:
look at qv2ray,  which should be considered the most comprehensive linux application

```cpp
#ifdef Q_OS_UNIX
    if (qEnvironmentVariableIsSet("APPIMAGE"))
        list << makeAbs(QCoreApplication::applicationDirPath() + "/../share/qv2ray/" + dirName);

    if (qEnvironmentVariableIsSet("SNAP"))
        list << makeAbs(qEnvironmentVariable("SNAP") + "/usr/share/qv2ray/" + dirName);

    if (qEnvironmentVariableIsSet("XDG_DATA_DIRS"))
        list << makeAbs(qEnvironmentVariable("XDG_DATA_DIRS") + "/" + dirName);

    list << makeAbs("/usr/local/share/qv2ray/" + dirName);
    list << makeAbs("/usr/local/lib/qv2ray/" + dirName);
    list << makeAbs("/usr/share/qv2ray/" + dirName);
    list << makeAbs("/usr/lib/qv2ray/" + dirName);
    list << makeAbs("/lib/qv2ray/" + dirName);
#endif

```

APPIMAGE use relative path,
SNAP Actually I have not used this format，seems the root path should be found from the SNAP environment variable


https://github.com/Qv2ray/Qv2ray/blob/2b9a5ecc239587517b34af123e4d04df8a7badf9/src/components/translations/QvTranslator.cpp#L11-L23
https://github.com/Qv2ray/Qv2ray/blob/2b9a5ecc239587517b34af123e4d04df8a7badf9/src/base/Qv2rayBaseApplication.cpp#L37-L77

dail8859:
Recently I took another attempt to see how best to load various translation files. I've come to the conclusion that trying to check so many possible locations is more of a pain than what it is worth currently.

I've merged #345 that embeds the translation files as QResources. If there ever become more/larger translation files then maybe it will be worth a deeper look then.

PR #345, opened 2023-03-17
Embed translation files

### LINKED ISSUE #155: Application translation files are not included in AppImage

dail8859:
Due to the various platform configurations (especially Linux) it makes it difficult to systematically locate the installed translation files. Since the number and size of the translation files are small, it makes more sense to embed these as qresources in the exe itself so loading the translation files can be deterministic.

Closes #155


dail8859:
> build\qrc\qmake_qmake_qm_files.qrc: Warning: potential duplicate alias detected: 'NotepadNext.qm'

Looking at the makefiles generated, there seems to be a bug with qmake using `embed_translations` for Qt5.

Since the translation files have a `.` in the file name it tries to create two `NotepadNext.qm` resource files rather than `NotepadNext.sv_SE.qm` and `NotepadNext.zh_CN.qm`. Using underscores in the `ts` files rather than `.` fixes the issue. Will push a change later today once I get time.

PR #363, opened 2023-04-12
Add to set filter in the folder dock 

KangLin:
The default folder is not show hide files. So, Add it!


dail8859:
Could you provide some additional details what issues this solves and what functionality it adds? and screenshots?

PR #366, opened 2023-04-13
Show hidden file in folder workspace

KangLin:
- Add a filter in the dock of the folder workspace
- Add the title bar

截图 2023-04-13 13-36-27
截图 2023-04-13 13-36-18


KangLin:
@dail8859 Please review.

dail8859:
@KangLin Thanks for the PR.

So the functionality change of toggling hidden files makes sense and worth exposing that kind of configuration.

Testing out the functionality seems to have some issues that it is *only* trying to show hidden files (`Filters=256`, which means only `QDir::Hidden` is the only flag set). But I'm sure that's an easy fix.

---

As far as the actual implementation...I like the idea of putting a small extra widget in the title bar but the custom one is a bit confusing with the custom icons and doesn't look like any of the others. 

image

And that much extra code doesn't seem to be worth it when it could be easily solved as a checkbox on the preferences dialog.

So some options to implement this...

- If it was as simple as just adding an extra widget to the already existing toolbar to keep the default look that might be acceptable...if it was determined that option needed to be easily accessible. I looked into trying to hook into the QDockWidget's current title bar but that does not seem possible as far as I can tell.
- As suggested it could easily be a checkbox in the preferences dialog.
- The other thing is probably at some point in the future the "Folder As Workspace" dock will probably need some additional buttons similar to Notepad++, so it may be worth just adding some sort of quick options button in there to toggle the option. But again I don't really see the need to make it easily accessible just to save the user a few clicks. See screenshot of Notepad++'s buttons
image


I'm certainly open to other suggestions but the CTitleBar dependency and the non-standard svgs I dont feel is the best direction currently.

KangLin:
> As far as the actual implementation...I like the idea of putting a small extra widget in the title bar but the custom one is a bit confusing with the custom icons and doesn't look like any of the others.

Yes. so that Refactor all.


> So some options to implement this...
>
>      If it was as simple as just adding an extra widget to the already existing toolbar to keep the default look that might be acceptable...if it was determined that option needed to be easily accessible. I looked into trying to hook into the QDockWidget's current title bar but that does not seem possible as far as I can tell.

That's why CTitleBar was added. and refactor all code using QDockwdiget with CTilteBar. Use styles to keep the same.

>      As suggested it could easily be a checkbox in the preferences dialog.

This feature is also required.

> 
>      The other thing is probably at some point in the future the "Folder As Workspace" dock will probably need some additional buttons similar to Notepad++, so it may be worth just adding some sort of quick options button in there to toggle the option. But again I don't really see the need to make it easily accessible just to save the user a few clicks. See screenshot of Notepad++'s buttons
>       image
> 

I don't recommend using the way. In that case, there will be fewer work areas.

> I'm certainly open to other suggestions but the CTitleBar dependency and the non-standard svgs I dont feel is the best direction currently.

QWidget's title bar is styled locally system, which may differ from the application's style. QT does not provide a solution to this problem. so that I add CTitleBar to resolve it.
SVG is non-standard, but it is not achieved with standard formats for scaling images. But Qt 5.10 and above are well supported.

dail8859:
For now I think the custom title bar is too much customization. Even if all the other dock widgets used it to keep the same look. Not saying something like this might not be useful in the future, but its best to start off simple and add functionality if it is required.

Think a simple checkbox in the preferences dialog is a perfect first step. Things can always be adjusted later if there is a solid need.

Issue #374, opened 2023-05-10
No response when scrolling up and down


czlitony:
## Problem
Opening a 50MB text file with Notepad Next, scrolling down and up, Notepad Next will become no response. Mac behaves much worse than Windows in this scenario.

## Analysis
From Mac sample, this problem is caused by `URLFinder::findURLs`. The problem can't be reproduced when `URLFinder` is disabled, so it's sure the issue is caused by `URLFinder::findURLs`. 
Mac sample:
<img width="1792" alt="image" src="https://github.com/dail8859/NotepadNext/assets/20141496/9faef90b-030c-4d45-9023-bd4686d07fb6">
[Sample of Notepad Next.txt](https://github.com/dail8859/NotepadNext/files/11441853/Sample.of.Notepad.Next.txt)

In `URLFinder::findURLs` there is a while-loop that is heavy. if `URLFinder::findURLs` is called frequently, then it will cause the no repsonse issue.
<img width="791" alt="image" src="https://github.com/dail8859/NotepadNext/assets/20141496/3c88afd2-e815-4f43-8a15-9838d1e59ca0">

## Comment
This issue is easy to reproduce on my Mac and Windows machines. Since scrolling down/up is the basic function, it's worth spending time to fix it with high priority.

czlitony:
Root cause is in 
```
    while(linesLeftToProcess >= 0 && currentLine < editor->lineCount()) {
        // Should only happen if the line is hidden
        if (!editor->lineVisible(currentLine)) {
            currentLine++;
            continue;
        }
```
When scrolling, many visible lines will become invisible, so the while loop will be executed many times.

czlitony:
I have not found a way to fix this issue perfectly. I'd like to remove the logic:
https://github.com/dail8859/NotepadNext/blob/cedeff71cdf171e3f70041df181039ad369df43d/src/NotepadNext/decorators/URLFinder.cpp#L113

@dail8859 what do you think of this change?

czlitony:
Disabling the Scintilla::Update::VScroll handling logic will make the URLFinder doesn't work normally for even small text files. I'll make a further investigation to fix this issue.

dail8859:
The intent is to just look at the text that is visible on the screen. After scrolling it would then "reprocess" the text in the screen. Since scroll events can happen in rapid succession, that's why I tried out delaying it a bit with a timer.

A guess may be poor performance with regex searching.

czlitony:
[...]

PR #377, opened 2023-05-13
Fix issue #374 No response when scrolling up and down

czlitony:
# Description
Fix issue  #374 No response when scrolling up and down.

# Analysis
Scintilla runs the regex search too slowly, so `while (editor->send(SCI_FINDTEXT, flags, (sptr_t)&ttf) != -1)` takes a long, then blocks the main thread.

# Fix
Now decorate the link texts in a line as follows:
1. Get line text
2. Get all the link texts in the line with `QRegExp`;
3. Iterator the link texts, use `editor->findText(SCFIND_MATCHCASE, ...)` to find the link text, then pass the fond link text's position to `editor->indicatorFillRange`

Before the fix:
reproduced_response_issue

After the fix: 
fixed_response_issue

### DIFF
--- a/src/NotepadNext/decorators/URLFinder.cpp
+++ b/src/NotepadNext/decorators/URLFinder.cpp
@@ -19,7 +19,7 @@
 #include <QDesktopServices>
 #include <QTimer>
 #include <QUrl>
-#include <QRegExp>
+#include <QRegularExpression>
 
 #include "URLFinder.h"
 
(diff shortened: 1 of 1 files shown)


dail8859:
Thanks for the PR. I've been quite busy so when I get some free time I'll take a deeper looks.

I think this demonstrates that the custom regex search implementation in https://github.com/dail8859/NotepadNext/blob/master/src/NotepadNext/QRegexSearch.cpp definitely isn't efficient. I'm assuming it is due to all the character counting that has to be done. Several other editors using Scintilla implement custom regex engines. I've stuck with `QRegularExpression` for now since that is built into Qt, but there I need to look at other implementations to see if they have similar issues. If it can't be made to work efficiently then stealing Notepad++'s regular expression code may be worth it since it is tried and tested for years.

I'm not against the PR but I'd fear that other parts of the code may be affected at some point in the future where regular expressions are used heavily in large files.

czlitony:
I didn't notify the implementation in `QRegexSearch::FindText`, it's better to check if `QRegexSearch::FindText` can be improved.

czlitony:
This fix is not the best solution, so closing it. I have opened another jira to fix this issue https://github.com/dail8859/NotepadNext/pull/378.

PR #378, opened 2023-05-14
Fix issue #374 No response when scrolling up and down

czlitony:
# Description
Fix issue  #374 No response when scrolling up and down.

# Analysis
Scintilla runs the regex search too slowly, so `while (editor->send(SCI_FINDTEXT, flags, (sptr_t)&ttf) != -1)` takes a long, then blocks the main thread.

# Fix
Before the fix:
reproduced_response_issue

After the fix: 
fixed_response_issue

### DIFF
--- a/src/NotepadNext/QRegexSearch.cpp
+++ b/src/NotepadNext/QRegexSearch.cpp
@@ -77,10 +77,11 @@ Sci::Position QRegexSearch::FindText(Document *doc, Sci::Position minPos, Sci::P
 
     // Get the bytes from the document. No need to go past maxPos bytes
     // Not actually sure if this copies the data or not
-    const QString utf8 = QString::fromUtf8(doc->BufferPointer(), maxPos);
+    const Sci::Position rangeLength = maxPos - minPos;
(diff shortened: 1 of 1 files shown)


src/NotepadNext/QRegexSearch.cpp
```diff
@@ -77,18 +77,19 @@ Sci::Position QRegexSearch::FindText(Document *doc, Sci::Position minPos, Sci::P
 
     // Get the bytes from the document. No need to go past maxPos bytes
     // Not actually sure if this copies the data or not
-    const QString utf8 = QString::fromUtf8(doc->BufferPointer(), maxPos);
+    const Sci::Position rangeLength = maxPos - minPos;
+    const QString utf8 = QString::fromUtf8(doc->RangePointer(minPos, rangeLength), rangeLength);
```
czlitony, line 81: Avoid creating the big string, then searching speed will be much faster.

czlitony:
With this change, the regex search is also faster.
image

dail8859:
Thanks for the PR! I did some testing to make sure it wasn't affected by 3- and 4-byte characters and everything seems to work great!

PR #379, opened 2023-05-17
Fix search result font size too small issue on Mac

czlitony:
# Problem

The search result font size is too small on Mac.
Snipaste_2023-05-17_20-56-14

# Fix
Increasing the search result font size to 14 on Mac.
Snipaste_2023-05-17_20-59-18


dail8859:
I'm not sure how high DPI and screen resolution come into play so I committed a slight variation on this in 41e48b724d2f1c50a2a4d4a4f9c523cc79830bb4 so it only affects MacOS.

I think in the future it may be worth making this a bit more dynamic and either pull the font size from the editor, or making the search results font size configurable or zoom in/out. But for now I think its best to keep changes simple. Thanks for the PR!

PR #380, opened 2023-05-17
Fix hex view font family issue

czlitony:
# Problem
When opening Notepad Next on Mac, can always see this warning:
```
[     0.595] W: Populating font family aliases took 147 ms. Replace uses of missing font family "Monospace" with one that exists to avoid this cost.
```

# Fix
Changing hex view font family to "Courier New"

### DIFF
--- a/src/NotepadNext/docks/HexViewerDock.cpp
+++ b/src/NotepadNext/docks/HexViewerDock.cpp
@@ -16,6 +16,7 @@
  * along with Notepad Next.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+#include <QFontDatabase>
 
 #include "MainWindow.h"
 #include "ScintillaNext.h"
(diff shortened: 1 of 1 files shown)


dail8859:
I did some digging as I'm not sure hard coding "Courier New" is the best option since it is not guaranteed to exist on systems, and it forces it to a specific font.

I ran across this same issue here: https://bugs.dolphin-emu.org/issues/12035#change-738847

Which was solved by this change:
https://github.com/dolphin-emu/dolphin/pull/9085/files

So I'm curious if...

```cpp
    QFont font = QFont(QFontDatabase::systemFont(QFontDatabase::FixedFont).family());
    ui->tblHexView->setFont(font);
```

...works on MacOS for you to eliminate the warning. This seems like it has more potential to accurately support the system's monospaced font as some systems configure it differently.

czlitony:
`QFont font = QFont(QFontDatabase::systemFont(QFontDatabase::FixedFont).family());` works.
image

PR #381, opened 2023-05-17
Increase search result column width

czlitony:
# Before
The search result column is stretched according to the app's main window width.
Snipaste_2023-05-17_21-55-52

# Now
Set the column width as 10000.
Snipaste_2023-05-17_21-56-58

### DIFF
--- a/src/NotepadNext/docks/SearchResultsDock.cpp
+++ b/src/NotepadNext/docks/SearchResultsDock.cpp
@@ -39,6 +39,8 @@ SearchResultsDock::SearchResultsDock(QWidget *parent) :
 {
     ui->setupUi(this);
 
+    ui->treeWidget->setColumnWidth(1, 10000);
+
     // Close the results when escape is pressed
     new QShortcut(QKeySequence::Cancel, this, this, &SearchResultsDock::close, Qt::WidgetWithChildrenShortcut);
 
--- a/src/NotepadNext/docks/SearchResultsDock.ui
+++ b/src/NotepadNext/docks/SearchResultsDock.ui
@@ -57,6 +57,9 @@
       <attribute name="headerVisible">
        <bool>false</bool>
       </attribute>
+      <attribute name="headerStretchLastSection">
+       <bool>false</bool>
+      </attribute>
       <column>
        <property name="text">
         <string notr="true">1</string>


dail8859:
@czlitony Greatly appreciate the PRs recently! Been a bit busy so I'll try to get around to reviewing them one by one as I have time.

> ui->treeWidget->setColumnWidth(1, 10000);

Usually when there is some arbitrary value like this it indicates some possible [code smell](https://en.wikipedia.org/wiki/Code_smell). This also has the downside of always resizing it to 10000 and forcing a scrollbar even when it is not needed.

I think a more dynamic solution is adding `ui->treeWidget->resizeColumnToContents(1);` at the end of `SearchResultsDock::completeSearch()` which I think will achieve the same results. I tested it quickly but not very thoroughly.

czlitony:
I'll try it.

czlitony:
Tried it, but doesn't work well:
image

dail8859:
Are you sure you added it at the bottom of `SearchResultsDock::completeSearch()`? It worked under windows for me.

There is another call to resize the column widths index 0 so I dont see why resizing column 1 wouldn't work.

```cpp
void SearchResultsDock::completeSearch()
{
    currentSearch = Q_NULLPTR;
    currentFile = Q_NULLPTR;
    currentFileCount = 0;
    totalFileHitCount = 0;
    totalHitCount = 0;
    ui->treeWidget->resizeColumnToContents(0);
    ui->treeWidget->resizeColumnToContents(1); // <-- resize text column here
}
```

czlitony:
Yes, I added it at the bottom of `SearchResultsDock::completeSearch()`, but didn't work on Mac. It worked on Windows.
`ui->treeWidget->header()->setSectionResizeMode(QHeaderView::ResizeToContents);` works for both Windows and Mac.

But resizing to content doesn't work well for the short lines, the search line and search file path line are truncated.
image

dail8859:
> But resizing to content doesn't work well for the short lines,

Your PR shows it has `headerStretchLastSection` set to false. Maybe changing to true helps?

czlitony:
With `ui->treeWidget->header()->setStretchLastSection(true);`, `ui->treeWidget->header()->setSectionResizeMode(QHeaderView::ResizeToContents);` doesn't work.

dail8859:
> Yes, I added it at the bottom of SearchResultsDock::completeSearch(), but didn't work on Mac. It worked on Windows.

This would indicate that it is a Qt bug most likely if it is not working on Mac. I'm still not a fan specifying a column with because it is arbitrary and also always forces a scrollbar to be shown all the time.

I just pushed 940cc193ea730d0c54ec120ca353850726cfd1d2 which behaves correctly on Windows. I think the best course of action is make Mac specific changes (e.g. using `ifdefs` to call `setColumnWidth(1, 10000);` or whatever is best) that way it gives better behavior even if it is not perfect for just Mac. I don't have a Mac to test out any of that on.

czlitony:
@dail8859 https://github.com/dail8859/NotepadNext/commit/940cc193ea730d0c54ec120ca353850726cfd1d2 also works on Mac, I don't know why it didn't work last time when I tried it, maybe something else changed by me will cause the problem.
Thanks for the fix, since this issue is resolved, I'll close this PR.

dail8859:
Great to hear!

czlitony:
@dail8859 https://github.com/dail8859/NotepadNext/commit/940cc193ea730d0c54ec120ca353850726cfd1d2 doesn't work well in this case that after a short line is matched, the previous search result's column will be shortened.
image

dail8859:
Good catch. Looks like it is just accounting for currently visible rows. There may be some better way to handle it.

PR #385, opened 2023-05-28
Fix bug SearchResultsDock doesn't resize column to contents

czlitony:
### Problem
Commit https://github.com/dail8859/NotepadNext/commit/940cc193ea730d0c54ec120ca353850726cfd1d2 optimized SearchResultsDock to make it resize the column to contents. However, it doesn't work well in this scenario that after a new short line is added in SearchResultsDock, the previous long line's column will be shortened.
image

### Fix
Call "ui->treeWidget->resizeColumnToContents(1)" to resize the column when the searched item is expanded.


dail8859:
Thanks!

PR #386, opened 2023-05-28
Collapse top level items when a new search result is added

czlitony:
### Problem
When a new search result is added, all the other search results will be collapsed, then always need to expand 2 times to check a previous search result.
original

### Fix
Only collapse top levels when a new search result is added.
optimized


dail8859:
Thanks for the PR and clear description of the issue. It makes reviewing the changes so much easier when the problem is easily understood!

PR #388, opened 2023-05-29
Turkish Language Support

CidQu:
Sorry, this take way longer than I anticipated.
I couldn't translate one thing because of how Turkish Grammer works, I left a comment at 158. Other than that, all translations were done with QT Linguist and are ready to merge!


dail8859:
@CidQu Awesome! Thanks!

PR #393, opened 2023-06-04
Optimize search results dock scroll on Mac

czlitony:
### Problem
In the search results dock, when hovering over the last line of the search results, the horizontal scrollbar will show and overlap with the last line, so it's hard to click the last line of the search results.
issue

### Fix
Change the search results dock's vertical scroll mode from the default value QAbstractItemView::ScrollPerPixel to QAbstractItemView::ScrollPerItem.
fix


dail8859:
Thanks!

PR #395, opened 2023-06-05
Implements a platform-dependent exit hotkey

data-man:


dail8859:
Hi @data-man thanks for the PR!

I know `Ctrl+Q` is much more common for Linux (and I'm assuming Mac). I took a look at `QKeySequence` and was hoping [QKeySequence::Quit](https://doc.qt.io/qt-6/qkeysequence.html#standard-shortcuts) would handle the cross-platform differences, but for some reason it doesn't provide a default on Windows.

However [Microsoft does specify](https://support.microsoft.com/en-us/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec#WindowsVersion=Windows_10) that `Alt+F4` should be used for quitting/exiting, so that seems a reasonable default. In the future I would like to have shortcut keys configurable, which would handle any situations where one would want it to be different.

So I think doing something similar to (untested):

```cpp
#ifdef Q_OS_WIN
    ui->actionExit->setShortcut(QKeySequence("Alt+F4"));
#else
    ui->actionExit->setShortcut(QKeySequence::Quit);
#endif
```

would be preferrable rather than setting it to `Ctrl+Q` for all platforms.

data-man:
> However [Microsoft does specify](https://support.microsoft.com/en-us/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec#WindowsVersion=Windows_10) that Alt+F4 should be used

This is a built-in action in any Windows application, without a programmer having to do anything, if I'm not mistaken. :)
Similar to what is done in KDE Plasma. 

> Alt + F4 | Close the active item, or exit the active app.

dail8859:
> This is a built-in action in any Windows application, without a programmer having to do anything

Correct. The only benefit of specifying it manually is so that it shows up in the menu as the assigned shortcut (which may or may not be beneficial to the user). So, to err on the side of simplicity, then yes you are correct and only `ui->actionExit->setShortcut(QKeySequence::Quit);` would be needed since it won't affect Windows, but the OS would still implement the shortcut.

data-man:
> ```c++
> #ifdef Q_OS_WIN
>    ui->actionExit->setShortcut(QKeySequence("Alt+F4"));
>#else
>    ui->actionExit->setShortcut(QKeySequence::Quit);
>#endif
>```

Hmm, [FocusWriter](https://github.com/gottcode/focuswriter) uses this:

```c++
window.cpp
   624:         m_document_cache_thread->quit();
  1311:         m_actions["Quit"] = file_menu->addAction(QIcon::fromTheme("application-exit"), tr("&Quit"), this, &Window::close);
  1312:         m_actions["Quit"]->setShortcut(keyBinding(QKeySequence::Quit, tr("Ctrl+Q")));
  1313:         m_actions["Quit"]->setMenuRole(QAction::QuitRole);
```

dail8859:
> m_actions["Quit"]->setShortcut(keyBinding(QKeySequence::Quit, tr("Ctrl+Q")));

`keyBinding()` seems to allow a fallback shortcut of `Ctrl+Q`. Given that it is a non-standard Microsoft shortcut, and also Notepad++ uses `Ctrl+Q` for commenting, I think using `Ctrl+Q` on Windows would only cause confusion to Windows users.

data-man:
> `ui->actionExit->setShortcut(QKeySequence::Quit);`

Ok, it's work for me in Linux! :)

Should `<string>Ctrl+Q</string>` be removed from `MainWindow.ui`?

dail8859:
> Should <string>Ctrl+Q</string> be removed from MainWindow.ui?

Yeah I think for clarity sake it can be removed to not cause any confusion later since it is being set outside of the ui file.

data-man:
It's ready, I hope.

dail8859:
Awesome. Thanks!

PR #406, opened 2023-06-24
Less + SCSS, new keywords/extensions for JS/HTML

Azq2:
New:
- SCSS support
- Less support
- Added .vue to html
- Added .mjs (es modules) / .cjs (commonjs modules) to js
- Add keywords `as` / `from` to js (example: `import a as b from 'module';`)


dail8859:
Awesome! The code changes look correct and very briefly tried some Less and SCSS examples. I'm not familiar with those languages at all but on the surface things seem to be working correctly.

PR #419, opened 2023-07-23
Fix tab's title/text elided when lots of files/tabs opened

wayne-zhang:


src/NotepadNext/DockedEditor.cpp
```diff
@@ -154,6 +158,9 @@ void DockedEditor::addEditor(ScintillaNext *editor)
     // Create the dock widget for the editor
     ads::CDockWidget *dockWidget = new ads::CDockWidget(editor->getName());
 
+    // Make tab fixed size so that tab's title/text will not be elided
+    dockWidget->tabWidget()->setSizePolicy(QSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed));
```
dail8859, line 162: Maybe try `CDockWidgetTab::setElideMode()`? That might remove the need to set `DockAreaDynamicTabsMenuButtonVisibility` to `false` above.

https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/blob/master/src/DockWidgetTab.h#L160

dail8859:
Appreciate the work on the PRs. I'm still hoping to get around to reviewing these _eventually_. Been busy with work and a new 👶

wayne-zhang:
No worries mate, it's our software and I should contribute not just use it :)

Tried to make tabs can be scrolled but it seems not possible unless patch/enhance ads. I have created a feature request in ads https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/issues/539.

If nobody will implement that feature, may be I can try.

src/NotepadNext/DockedEditor.cpp
wayne-zhang: Tested, unfortunately it doesn't work.

setElide(Qt::ElideNone) has the same effect as set fixed size policy, so use elide mode to disable file name elide. However, it set elide mode will not make dynamic menu bar woks - menu bar still not shown when more files/tabs opened than view tab. 

So make menu bar can be used to select files (when more files opened than view port), dynamic menu bar have to turn off. Possibly this is a bug of ads.

dail8859:
Sorry for the super long delay on this! I checked it out a bit and I think what you have is the best way for now. Part of the confusion I was running into was a bug that appears to be ADS related. I'll need to spend some more time to create a reproducible test case and contact that project.

PR #455, opened 2023-10-01
Add zh_TW Traditional Chinese locale

PeterDaveHello:
It's an early but also hopefully working version of Traditional Chinese translation 🎉


dail8859:
Awesome! Thanks so much!

PR #463, opened 2023-10-10
Build with Qt6.6

dail8859:


.github/workflows/build.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         - 
           qt_version: "6.5"
           modules: "qt5compat"
+        - 
+          qt_version: "6.6"
```
elcuco, line 31: I would kill everything but 5.15 and 6.6.1. I am unsure about 5.15

PR #494, opened 2023-12-22
Add note about MacOS font smoothing.

commonbreed:


dail8859:
Thanks for the PR! Sorry it took me so long to get to.

Changes have been merged with some minor tweaks.

In the future its best to keep two separate changes as different pull requests.

For the Mac font smoothing, if you get a chance can you post two images showing the difference? That way I'll know what to look for if someone reports an issue. Then I'll know how to fix it :)

Issue #506, opened 2024-02-15
Canceling "Copy As..." shows error


dail8859:
Select `File` > `Save a Copy As...` then click `Cancel`.

An error message is shown.

![image](https://github.com/dail8859/NotepadNext/assets/3694843/5157fb62-d1de-48dd-afa2-6e7cde300df6)

PR #507, opened 2024-02-19
return from writeToDisk function if the path given to the function is…

PratyakshSinha:
Return  QFileDevice::NoError from function writeToDisk if the path provided to is empty the error was because the file.open was used even in case of empty file path.


dail8859:
Can you provide more information as to what this PR is fixing?

My gut feeling this is hiding some other problem. Logically this shouldn't ever happen that `path` is empty. _If anything_, this should be an assertion failure or throw an exception.

dail8859:
Oh. I see this is related to #506.

Catching the problem at the lowest level isn't a good idea. Take a look at some of the other functions in `MainWindow.cpp`. There are a couple other calls to `FileDialogHelpers::getSaveFileName()` that might be a good reference.

PratyakshSinha:
First of all sorry for not mentioning clearly that this PR is fixing issue 506 I'm new to open source development. Although I mentioned it in the branch name but I should've mentioned in details also. 

As for your comment sure I'll look into it right now and try fixing it.

PR #521, opened 2024-03-03
Added Russian translation.

dchapyshev:


dail8859:
Thanks for the PR! I'm curious about `qtbase_ru`. This makes me wonder if `qtbase_XXX` needs included for other languages. Do you happen to know what parts of the application this translates?

dchapyshev:
I think you can navigate by file size. For example, for the Russian language, file qt_ru.ts is empty, and file qtbase_ru.ts contains translations (but there are languages where this is the other way around). To be on the safe side, you can include both files to have a universal approach.
image

dail8859:
That is very helpful to know. I'm not sure how all Qt loads translations. This application is only looking for `qt_XXX` and not `qtbase_XXX`

https://github.com/dail8859/NotepadNext/blob/cdb2b84db09409299ed02e7043156b391e2ad8ac/src/NotepadNext/TranslationManager.cpp#L85

According to the [`qt_ru.ts`](https://github.com/qt/qttranslations/blob/dev/translations/qt_ru.ts) file it has a dependency on `qtbase_ru`

I'm not sure if this means when it loads `qt_ru` it will find the `qtbase_ru` or what. This may mean `qt_ru` also needs included even though it is mostly empty, that may satisfy the need to load the `qtbase` translation.

I'll need to do some testing and reading up to make sure all possible translations are included and properly loaded.

dchapyshev:
Files qt_ru and qtbase_ru must be loaded separately. When you load one of the files, the other will not automatically load.
I have already done loading translations in [my project](https://github.com/dchapyshev/aspia/blob/master/source/qt_base/locale_loader.cc). This code has an example of universal language loading, you could use a similar approach. Maybe this will help you.

dail8859:
> I have already done loading translations in [my project](https://github.com/dchapyshev/aspia/blob/master/source/qt_base/locale_loader.cc).

Oh awesome! Thanks!

I'll go ahead and merge this knowing that it won't actually be loading the qtbase file yet. I'll look into adding that in the future soon.

dail8859:
Resolved conflict. Once the CI checks pass, I'll merge this in.

Thanks!

PR #527, opened 2024-03-28
Check filename before save on Save a Copy As...

### LINKED ISSUE #506: Canceling "Copy As..." shows error

detinho:
Closes #506

Making the same check done on `MainWindow::saveCurrentFileAsDialog`


dail8859:
Awesome! Thanks! 🎉

PR #529, opened 2024-03-28
CFBundleIdentifier

FnControlOption:
Updated `CFBundleIdentifier` to follow the format used by some other OSS projects such as [OpenToonz](https://github.com/opentoonz/opentoonz): `io.github.opentoonz.OpenToonz`

Old: `~/Library/Saved Application State/com.yourcompany.NotepadNext.savedState` 

New: `~/Library/Saved Application State/io.github.dail8859.NotepadNext.savedState`

### DIFF
--- a/deploy/macos/info.plist
+++ b/deploy/macos/info.plist
@@ -9,7 +9,7 @@
 	<key>CFBundleIconFile</key>
 	<string>NotepadNext.icns</string>
 	<key>CFBundleIdentifier</key>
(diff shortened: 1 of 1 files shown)


dail8859:
I'm not a Mac user at all so I'm always open to suggestions.  And this definitely seems more reasonable than what was there.

Out of curiosity how does this affect the user. Do they see this value or does the OS use it in some way I'm assuming?

PR #538, opened 2024-04-03
Adapt the Find/Replace dialog UI for MAC

ErickWei:
Before:
image
After:
image


dail8859:
Everything seems fine still for Windows and obviously the layout looks way better for MacOS!

Thanks!

PR #549, opened 2024-04-22
Update Linux screenshot

### LINKED ISSUE #545: Better Linux Screenshot
It would be great if some kind soul could update the Linux screenshot to something more polished than what is currently there.

[doc/screenshot_linux.png](https://github.com/dail8859/NotepadNext/blob/master/doc/screenshot_linux.png)

If it was similar to the Windows one, that would be even better :)

[doc/screenshot.png](https://github.com/dail8859/NotepadNext/blob/master/doc/screenshot.png)

japanoise:
Closes #545


dail8859:
I know it might be kind of nit-picky but probably best to match the same image size. This image is what shows up in some other places (like [flathub](https://flathub.org/apps/com.github.dail8859.NotepadNext)) so I'm not sure how particular it is about the image size.

While you are at it, maybe one other suggestion if you don't mind, can you select a word in one of the views so it shows the green highlighting? That way the green markers show up on the scroll bar. I like showing that off since not all editors have that feature.

Appreciate the help so far! :)

japanoise:
> I know it might be kind of nit-picky but probably best to match the same image size. This image is what shows up in some other places (like [flathub](https://flathub.org/apps/com.github.dail8859.NotepadNext)) so I'm not sure how particular it is about the image size.
> 
> While you are at it, maybe one other suggestion if you don't mind, can you select a word in one of the views so it shows the green highlighting? That way the green markers show up on the scroll bar. I like showing that off since not all editors have that feature.
> 
> Appreciate the help so far! :)

Sure thing, can do. I have the fork up & the scaffolding in place. I'm working today so it won't be until this evening, but I'll update the screenshot for you when I get in. I can also fix the annoying problem with the titlebar.

dail8859:
Great! No rush at all 👍

dail8859:
Awesome! Thanks!

PR #556, opened 2024-05-07
Context Menu operations on Folder as Workspace tree

SimLV:
For now:

- New Folder (for folders and root)
- Delete
- Rename
- Save here (save active document under a new name)

### DIFF
--- a/src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
+++ b/src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
@@ -32,6 +32,7 @@ FolderAsWorkspaceDock::FolderAsWorkspaceDock(QWidget *parent) :
 {
     ui->setupUi(this);
 
+    model->setReadOnly(false);
     ui->treeView->setModel(model);
     ui->treeView->header()->hideSection(1);
     ui->treeView->header()->hideSection(2);
@@ -42,6 +43,7 @@ FolderAsWorkspaceDock::FolderAsWorkspaceDock(QWidget *parent) :
             emit fileDoubleClicked(model->filePath(index));
         }
     });
+    connect(ui->treeView, &QTreeView::customContextMenuRequested, this, &FolderAsWorkspaceDock::onCustomContextMenu);
 
     ApplicationSettings settings;
     setRootPath(settings.get(rootPathSetting));
@@ -64,4 +66,42 @@ void FolderAsWorkspaceDock::setRootPath(const QString dir)
 QString FolderAsWorkspaceDock::rootPath() const
 {
     return model->rootPath();
+}
+
+void FolderAsWorkspaceDock::onCustomContextMenu(const QPoint &point)
+{
+    QModelIndex index = ui->treeView->indexAt(point);
+    if (!index.isValid()) {
+        lastSelectedItem = model->index(0, 0);
+        ui->menuEmpty->exec(ui->treeView->viewport()->mapToGlobal(point));
+        return;
+    }
+    lastSelectedItem = index;
+    if (model->isDir(index)) {
+        ui->menuDirectory->exec(ui->treeView->viewport()->mapToGlobal(point));
+    }
+    else {
+        ui->menuFile->exec(ui->treeView->viewport()->mapToGlobal(point));
+    }
+}
+
+void FolderAsWorkspaceDock::on_actionNewFile_triggered()
+{
+    qInfo(Q_FUNC_INFO);
+}
+
+void FolderAsWorkspaceDock::on_actionNewFolder_triggered()
+{
+    qInfo(Q_FUNC_INFO);
+}
+
+void FolderAsWorkspaceDock::on_actionRename_triggered()
+{
+    ui->treeView->setCurrentIndex(lastSelectedItem);
+    ui->treeView->edit(lastSelectedItem);
+}
+
+void FolderAsWorkspaceDock::on_actionDelete_triggered()
+{
+    qInfo(Q_FUNC_INFO);
 }
--- a/src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
+++ b/src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
@@ -18,18 +18,24 @@
 
 
 #include "FolderAsWorkspaceDock.h"
-#include "ApplicationSettings.h"
-#include "ui_FolderAsWorkspaceDock.h"
 
 #include <QFileSystemModel>
+#include <QMessageBox>
+
+#include "ApplicationSettings.h"
+#include "MainWindow.h"
+#include "ui_FolderAsWorkspaceDock.h"
+
+QString NEW_DIR_TEMPLATE("dir_%1");
 
 ApplicationSetting<QString> rootPathSetting{"FolderAsWorkspace/RootPath"};
 
-FolderAsWorkspaceDock::FolderAsWorkspaceDock(QWidget *parent) :
+FolderAsWorkspaceDock::FolderAsWorkspaceDock(MainWindow *parent) :
     QDockWidget(parent),
-    ui(new Ui::FolderAsWorkspaceDock),
-    model(new QFileSystemModel(this))
+    window(parent)
 {
+    ui = new Ui::FolderAsWorkspaceDock;
+    model = new QFileSystemModel(this);
     ui->setupUi(this);
 
     model->setReadOnly(false);
@@ -72,7 +78,7 @@ void FolderAsWorkspaceDock::onCustomContextMenu(const QPoint &point)
 {
     QModelIndex index = ui->treeView->indexAt(point);
     if (!index.isValid()) {
-        lastSelectedItem = model->index(0, 0);
+        lastSelectedItem = model->index(rootPath());
         ui->menuEmpty->exec(ui->treeView->viewport()->mapToGlobal(point));
         return;
     }
@@ -85,14 +91,45 @@ void FolderAsWorkspaceDock::onCustomContextMenu(const QPoint &point)
     }
 }
 
-void FolderAsWorkspaceDock::on_actionNewFile_triggered()
+void FolderAsWorkspaceDock::on_actionSaveHere_triggered()
 {
-    qInfo(Q_FUNC_INFO);
+    QDir parentDir(model->filePath(lastSelectedItem));
+    auto doc = window->currentEditor();
+    QString dstName(parentDir.absoluteFilePath(doc->getName()));
+
+    if (doc->saveAs(dstName) != QFileDevice::NoError)
+    {
+        qWarning("Unable to save %s", dstName.toUtf8().constData());
+    }
+    else
+    {
+        auto newItem = model->index(dstName);
+        if (!doc->isFile()) {
+            doc->setFileInfo(dstName);
+        }
+        ui->treeView->setCurrentIndex(newItem);
+        ui->treeView->edit(newItem);
+    }
 }
 
 void FolderAsWorkspaceDock::on_actionNewFolder_triggered()
 {
-    qInfo(Q_FUNC_INFO);
+    QDir parentDir(model->filePath(lastSelectedItem));
+    int i = 1;
+
+    for (;parentDir.exists(NEW_DIR_TEMPLATE.arg(i)); i++) {
+    // Intentional
+    }
+    QString dstName = NEW_DIR_TEMPLATE.arg(i);
+
+    auto newItem = model->mkdir(lastSelectedItem, dstName);
+    if (!newItem.isValid()) {
+        qWarning("Unable to create %s", dstName.toUtf8().constData());
+    }
+    else {
+        ui->treeView->setCurrentIndex(newItem);
+        ui->treeView->edit(newItem);
+    }
 }
 
 void FolderAsWorkspaceDock::on_actionRename_triggered()
@@ -103,5 +140,22 @@ void FolderAsWorkspaceDock::on_actionRename_triggered()
 
 void FolderAsWorkspaceDock::on_actionDelete_triggered()
 {
-    qInfo(Q_FUNC_INFO);
+    bool ret;
+    QString path(model->filePath(lastSelectedItem));
+    QMessageBox::StandardButton reply = QMessageBox::question(this, tr("Delete Item"),
+        tr("Are you sure you want to delete <b>%1</b>?").arg(path));
+
+    if (reply == QMessageBox::Yes)
+    {
+        if (model->isDir(lastSelectedItem)) {
+            ret = model->rmdir(lastSelectedItem);
+        }
+        else {
+            ret = model->remove(lastSelectedItem);
+        }
+        if (!ret)
+        {
+            qWarning("Unable to delete %s", path.toUtf8().constData());
+        }
+    }
 }
--- a/src/NotepadNext/docks/FolderAsWorkspaceDock.h
+++ b/src/NotepadNext/docks/FolderAsWorkspaceDock.h
@@ -21,6 +21,7 @@
 #define FOLDERASWORKSPACEDOCK_H
 
 #include <QDockWidget>
+#include <QModelIndex>
 
 namespace Ui {
 class FolderAsWorkspaceDock;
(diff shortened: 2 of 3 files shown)


src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
```diff
@@ -18,20 +18,27 @@
...
 
-FolderAsWorkspaceDock::FolderAsWorkspaceDock(QWidget *parent) :
+FolderAsWorkspaceDock::FolderAsWorkspaceDock(MainWindow *parent) :
     QDockWidget(parent),
-    ui(new Ui::FolderAsWorkspaceDock),
-    model(new QFileSystemModel(this))
+    window(parent)
 {
+    ui = new Ui::FolderAsWorkspaceDock;
+    model = new QFileSystemModel(this);
```
dail8859, line 38: Any particular reason these couldn't be kept in the initializer list?

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
```diff
@@ -18,20 +18,27 @@
...
+
+#include <QFileSystemModel>
+#include <QMessageBox>
+
 #include "ApplicationSettings.h"
+#include "MainWindow.h"
 #include "ui_FolderAsWorkspaceDock.h"
 
-#include <QFileSystemModel>
+QString NEW_DIR_TEMPLATE("dir_%1");
```
dail8859, line 29: I'd have to look deeper how this is used but might make sense to keep it inside the class so this can be translated.

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
```diff
@@ -65,3 +74,95 @@ QString FolderAsWorkspaceDock::rootPath() const
...
+    }
+}
+
+void FolderAsWorkspaceDock::on_actionSaveHere_triggered()
+{
+    QDir parentDir(model->filePath(lastSelectedItem));
+    auto doc = window->currentEditor();
+    QString dstName(parentDir.absoluteFilePath(doc->getName()));
+
+    if (doc->saveAs(dstName) != QFileDevice::NoError)
```
dail8859, line 101: In most cases I prefer the "happy path" to be the first part (e.g. switching this to `==`). Just helps with readability since you have to skip over the failure case and look at the `else` clause to see the natural flow.

Also, this is a bit of a code 'smell' that the FolderAsWorkspace is messing with the lower editor API. Might be possible to use something like `window->saveFileAs(...)` so that incase there are errors it will be handled the same way as a normal file would if it failed.

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
```diff
@@ -65,3 +74,95 @@ QString FolderAsWorkspaceDock::rootPath() const
...
+    }
+    else {
+        ui->menuFile->exec(ui->treeView->viewport()->mapToGlobal(point));
+    }
+}
+
+void FolderAsWorkspaceDock::on_actionSaveHere_triggered()
+{
+    QDir parentDir(model->filePath(lastSelectedItem));
+    auto doc = window->currentEditor();
```
dail8859, line 98: To keep consistent with the rest of the code base, rename `doc` to `editor`.

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
```diff
@@ -65,3 +74,95 @@ QString FolderAsWorkspaceDock::rootPath() const
...
+    QDir parentDir(model->filePath(lastSelectedItem));
+    int i = 1;
+
+    for (;parentDir.exists(NEW_DIR_TEMPLATE.arg(i)); i++) {
+    // Intentional
+    }
+    QString dstName = NEW_DIR_TEMPLATE.arg(i);
+
+    auto newItem = model->mkdir(lastSelectedItem, dstName);
+    if (!newItem.isValid()) {
```
dail8859, line 127: Switch this condition as well.

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
```diff
@@ -65,3 +74,95 @@ QString FolderAsWorkspaceDock::rootPath() const
...
+{
+    ui->treeView->setCurrentIndex(lastSelectedItem);
+    ui->treeView->edit(lastSelectedItem);
+}
+
+void FolderAsWorkspaceDock::onFileRenamed(const QString &path, const QString &oldName, const QString &newName)
+{
+    QDir dir(path);
+    QString fileName = dir.absoluteFilePath(oldName);
+    for(auto &&editor : window->editors())
```
dail8859, line 146: Might also leverage something in `window`...might need something new like `renameFile` to take an editor.

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
```diff
@@ -65,3 +74,95 @@ QString FolderAsWorkspaceDock::rootPath() const
...
+        {
+            editor->setName(newName);
+        }
+    }
+}
+
+void FolderAsWorkspaceDock::on_actionDelete_triggered()
+{
+    QString path(model->filePath(lastSelectedItem));
+    QMessageBox::StandardButton reply = QMessageBox::question(this, tr("Delete Item"),
```
dail8859, line 158: Probably use `window->moveFileToTrash()` here.

src/NotepadNext/docks/FolderAsWorkspaceDock.h
```diff
@@ -42,10 +44,22 @@ class FolderAsWorkspaceDock : public QDockWidget
 signals:
     void fileDoubleClicked(const QString &filePath);
 
+private slots:
+    void on_actionSaveHere_triggered();
+    void on_actionNewFolder_triggered();
+    void on_actionRename_triggered();
+    void on_actionDelete_triggered();
```
dail8859, line 51: Qt doesn't recommend the auto connect slots and I haven't used them elsewhere throughout the codebase, but for now it is fine. It can be cleaned up later.

src/NotepadNext/docks/FolderAsWorkspaceDock.ui
```diff
@@ -27,19 +27,77 @@
     <property name="bottomMargin">
      <number>0</number>
     </property>
+    <item>
+     <widget class="QMenu" name="menuFile">
```
dail8859, line 31: Can QtCreator create a QMenu in the ui file? or some other method (e.g. manually)? I'm just interested since I've never seen this done before :)

dail8859 (changes requested):
Appreciate the PR! I knew this functionality was definitely needed. I think this is a great first start. I'm sure there will be more added to the menu eventually.

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
SimLV: There was warning about `-Wreorder` I don't get how to remove

src/NotepadNext/docks/FolderAsWorkspaceDock.ui
SimLV: It can't. I googled for it but at least Qt 5 can't. For me it looks like minor inconvenience everyone accustomed to.

src/NotepadNext/docks/FolderAsWorkspaceDock.h
SimLV: M.. why/where? `connectSlotsByName` has no such notices at any version.

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
SimLV: There is some mess between `MainWindow` <-> `EditorManager` and also `ScintillaNext` as a view vs some nonexistent `Content` which could be file or i.e. temporary buffer

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
SimLV: done

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
SimLV: I separate UI questions inside MainWindow and use them here instead.

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
SimLV: done

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
SimLV: Moved parts into `MainWindow`

src/NotepadNext/docks/FolderAsWorkspaceDock.cpp
SimLV: Implemented both options but I have no windows box nearby to test on windows.

src/NotepadNext/docks/FolderAsWorkspaceDock.h
dail8859: clazy is reporting it:

image


https://github.com/KDE/clazy/blob/1.11/docs/checks/README-connect-by-name.md

dail8859:
I'll take a crack at this and push any updates. I think I might have been confused a bit and told you wrong in places. I guess I was thinking the "File List" window...which means there was always an `editor` object associated with it. But realize now for the Folder as Workspace it is more direct file system modification potentially.

dail8859:
I pushed updates to this branch. I made a quick first round of cleanups...hopefully I didnt break anything that was working previously.

Some things to note:
- I removed the permeant deletion for now. Think its better to just give the user an option move it to the trash. (if someone really wants it maybe it is a configuration later)
- I understand the intent of `closeByPath()`...the difficult part is trying to half use the FaW as an explorer window that also keeps the opened files in sync (renamed, deleted, etc)...which is part of the reason why I didn't tackle this in the first place :) Notepad++ doesn't provide those file level operations so it never has to worry about that. For example:
image
- It might be better to leverage the `EditorManager` to future proof some things when finding an editor by the file path, because in the future if there are multipe `window`s it is resposinble for coordinating the editors between them. Plus if an editor is opened multiple times etc.

Sorry...late night brain dump :)

src/NotepadNext/dialogs/MainWindow.cpp
```diff
@@ -1369,7 +1356,10 @@ void MainWindow::moveFileToTrash(ScintillaNext *editor)
 
     if (askMoveToTrash(filePath)) {
         if (editor->moveToTrash()) {
-            closeByPath(filePath);
+            closeCurrentFile();
```
SimLV, line 1359: Why are you closing current file? One could open one file and delete another one.

SimLV:
- I was searching for permanent deletion solution like in file explorers (Option+Delete means permanent delete) but this needs subclassing of QTreeView. 
- "Eat your own dogfood". At least minor file manipulations (new/rename/delete) are useful for me.
- I am working+thinking on "Run..." stuff but this is more complex

Do you need some other rework for this branch from me for now?

src/NotepadNext/docks/FolderAsWorkspaceDock.h
SimLV: It is good scripting usage to automate creation of all these boilerplate as a lua script inside NotepadNext.
For some future version obviously
```lua
local luaxml = require 'LuaXML'

local file_name = nn.activeEditor.path:gsub('.ui$')
local cpp_editor = nn.findEditorByPath(file_name .. '.cpp')
local h_editor = nn.findEditorByPath(file_name .. '.h')

assert(cpp_editor and h_editor, 'Please open corresponding .cpp and .h')
local cpp_bookmark = cpp_editor:findNamedBookmark('ctor signals')
local h_bookmark = cpp_editor:findNamedBookmark('slots')
assert(cpp_bookmark and h_bookmark, 'Please set "ctor signals" and "slots" bookmarks')

local items = {}
local tree = xml.load(file_name .. '.ui')
tree:iterate(function(el)
  table.insert(items, el.tag('name'))
end, 'action', nil, nil, true)

local class_name = find_class_name(cpp_bookmark) -- TODO

for i,name in ipairs(items) do
  local method_name = 'on'.. name:gsub('^[a-z]', function(s) return s:upper() end) .. 'Triggered'
  local full_method = 'void '..method_name..'()'
  if not h_editor:findText(full_method) then
   h_bookmark:appendText(full_method .. ';\n');
   cpp_bookmark:append('connect(ui->'..name..', &QAction::triggered, this, &'.. class_name.. '::'..method_name..');')
   cpp_editor:append(full_method..'\n{\n}\n')
  end
end
```

src/NotepadNext/dialogs/MainWindow.cpp
dail8859: You are right! Good catch! 😄 This is what I get for trying to modify it on little sleep.

dail8859:
> I was searching for permanent deletion solution like in file explorers (Option+Delete means permanent delete) but this needs subclassing of QTreeView.

Yeah I think if something isn't obvious it is good to err on the side of caution. The user needs to accept the prompt anyways so best to move it to the trash. Again this could be an app setting or something in the future but unless there is a real "need" I don't think it is needed. If you need to delete something "permanently" I think it would be best not to use a text editor's file management window :)

> "Eat your own dogfood". At least minor file manipulations (new/rename/delete) are useful for me.

Agreed. They are very nice and normally 'expected' features if you are going to be showing a list of files.

> I am working+thinking on "Run..." stuff but this is more complex

I'm not saying this is needed for now. I was just showing an example of what Notepad++ does. This stuff can always be added in later if there is a real use-case for it.

> Do you need some other rework for this branch from me for now?

No, I don't believe so currently. I need to find some more time to just sit down with it and work through some of this.

src/NotepadNext/docks/FolderAsWorkspaceDock.h
dail8859: Interesting idea, though I don't see Lua as a long-term solution. It was originally just added for debugging/testing purposes

SimLV: I see lua as very good "Glue language" to customize and script stuff. I am going to use NotepadNext as very customizable light editor (instead of NotepadQQ and ZeroBraneStudio) on projects where I don't need big IDE like PyCharm

dail8859:
So I had some time to step back and think about this. There is definitely more complexity that has to be handled to implement it correctly.

If you rename a folder, now you have to check *every file* in that folder as a possible candidate that is opened for editing because it's file path could have changed.

Same with deleting a folder. You now have to check *every file*.

I know the application doesn't gracefully handle when files change outside of the application yet, but there could be race conditions that the application sees the file change on disk and notify the user the file is missing/edited before the FaW logic could tell the application it is gone/edited.

Also there's a case that you delete the file, but the file changes weren't saved, then I guess you just close the file anyways? (not looking for a direct answer, just thinking out loud).

Most of these cases are possible to handle but could run into cases that a folder could contain hundreds or thousands of files and make the entire situation worse.

---

Remembering now, all this was why long ago I never added a context menu since it is desirable to have these kinds of file operations but hard to handle gracefully.

SimLV:
Some of such stuff could be fixed using timer and check for all opened files once a second.
Going to fix:
- [ ] Check files on Rename folder
   - all Editors will be renamed
- [ ] Check files on Delete folder
   - saved Editors will be closed
   - unsaved Editors will be kept open

dail8859:
> Some of such stuff could be fixed using timer and check for all opened files once a second.

Honestly that definitely doesn't not sound like a good idea and timers are usually just band aid for race conditions.

SimLV:
Then QFileSystemWatcher for each Editor bound to a file

SimLV:
@dail8859  could you please check and possible merge this PR. 
I think further improvements could be in other PRs.

dail8859:
I want to find some more time to review this again but not sure when that will be yet to identify any additional issues that need addressed or noted for future PRs. 

I'll be the first to admit I'm quite particular about code that gets merged and want to make sure. I do appreciate the time and effort you and others contribute so I'll do my best to set aside some time for this.

dail8859:
I did pick this back up recently and started hacking away at it. In the end I might simplify it down for now to get alot of the changes merged in and can tackle some of the technically challenging functionality later to narrow down testing it.

Agrament (approved):

Agrament (approved):

SimLV:
Merged current master and tested

Neustradamus:
@dail8859: Any news on this PR?

SimLV:
Merged current master

PR #590, opened 2024-07-03
Issue 325 - location of search dialog

pydavem:
Adds an option in the preferences dialog to locate the search dialog in the main window when opened, or open it in the last position it was opened at.


dail8859:
Thanks for the PR! It looks like the relevant changes look pretty good. The PR just needs some cleanup first for further review.

Issue #224, opened 2022-10-09
[Feature request] Copy Search results 


xkarub:
I am unable to copy the search results window. 
<img width="973" alt="image" src="https://user-images.githubusercontent.com/30411064/194765428-1cb72404-0652-40aa-b0fe-6f16530859d6.png">

It was a handy feature of notepad++. 
Can you please add it?

xkarub:
Hi, @dail8859 , is this going to be fixed with the commit that you referenced? Could we expect a release?
This is such a handy feature.
Thank you!

dail8859:
Sorry, that commit referenced the wrong issue.

I think the feature is definitely needed but not sure when I'll get around to it.

Would you care to describe exactly how you copy the search results in Notepad++? If I recall correctly there are several ways to do it. Just curious what your actual need is so it could be accurately addressed.

xkarub:
The main intention of using it, is after I find all results, I copy them to a new page and analyze them do other text operations. 

![how to use notepadd copy all results](https://user-images.githubusercontent.com/30411064/198876071-e4da4179-5785-477b-bb31-c7cb6cd25f90.gif)

xkarub:
Hi Dail,
Hope you got my message. Apprecaite you looking into it!

PR #596, opened 2024-07-05
[Feature request] Copy Search results #224

pydavem:
Adds a button to the Search Results dock to copy the contents to the clipboard

### DIFF
--- a/src/NotepadNext/docks/SearchResultsDock.cpp
+++ b/src/NotepadNext/docks/SearchResultsDock.cpp
@@ -51,6 +51,7 @@ SearchResultsDock::SearchResultsDock(QWidget *parent) :
 
     connect(ui->treeWidget, &QTreeWidget::itemActivated, this, &SearchResultsDock::itemActivated);
     connect(ui->treeWidget, &QTreeWidget::itemExpanded, this, &SearchResultsDock::itemExpanded);
+    connect(ui->btnCopyResults, &QPushButton::released,this, &SearchResultsDock::copySearchResultsToClipboard);
 
     connect(ui->treeWidget, &QTreeWidget::customContextMenuRequested, this, [=](const QPoint &pos) {
         QTreeWidgetItem *item = ui->treeWidget->itemAt(pos);
@@ -212,7 +213,7 @@ void SearchResultsDock::updateSearchStatus()
         currentFile->setText(0, QStringLiteral("%1 (%L2 hits)").arg(currentFilePath).arg(totalFileHitCount));
 }
 
-void SearchResultsDock::on_pb_copyResults_released()
+void SearchResultsDock::copySearchResultsToClipboard()
 {
     QClipboard *clipboard = QGuiApplication::clipboard();
     QString string;
--- a/src/NotepadNext/docks/SearchResultsDock.cpp
+++ b/src/NotepadNext/docks/SearchResultsDock.cpp
@@ -215,15 +215,15 @@ void SearchResultsDock::updateSearchStatus()
 
 void SearchResultsDock::copySearchResultsToClipboard()
 {
-    QClipboard *clipboard = QGuiApplication::clipboard();
-    QString string;
-
+    QStringList results;
     QTreeWidgetItemIterator it(ui->treeWidget);
+
     while (*it) {
-        string.append((*it)->text(0) + ' ' + (*it)->text(1) + '\n');
+        const QTreeWidgetItem *item = *it;
+        results.append(QStringLiteral("%1 %2").arg(item->text(0)).arg(item->text(1)));
         ++it;
     }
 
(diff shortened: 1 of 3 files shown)


src/NotepadNext/docks/SearchResultsDock.cpp
```diff
@@ -210,3 +211,18 @@ void SearchResultsDock::updateSearchStatus()
...
 }
+
+void SearchResultsDock::on_pb_copyResults_released()
+{
+    QClipboard *clipboard = QGuiApplication::clipboard();
+    QString string;
+
+    QTreeWidgetItemIterator it(ui->treeWidget);
+    while (*it) {
+        string.append((*it)->text(0) + ' ' + (*it)->text(1) + '\n');
```
dail8859, line 222: Should `\r\n` be used on Windows? Not sure if it makes a difference. Maybe the clipboard normalizes line ending.

dail8859, line 222: Is there a more efficient way to build this string? In theory the treeWidget might contain thousands of items. Maybe it is a non-concern, just curious is all. Maybe even something like QString().arg() etc would be more efficient?

I'm not sure.  Haven't looked into this myself yet. Just curious is all.

src/NotepadNext/docks/SearchResultsDock.h
```diff
@@ -54,6 +54,8 @@ private slots:
     void itemActivated(QTreeWidgetItem *item, int column);
     void itemExpanded(QTreeWidgetItem *item);
 
+    void on_pb_copyResults_released();
```
dail8859, line 57: I'm not a fan of using automatic connections. Qt also suggests not doing this any more. Renaming an object would break the callback without any compile time error.

Specifically using `connect(...)` in the constructor is more verbose but is safer since it would fail at compile time.

src/NotepadNext/docks/SearchResultsDock.ui
```diff
@@ -30,6 +30,13 @@
     <property name="bottomMargin">
      <number>0</number>
     </property>
+    <item>
+     <widget class="QPushButton" name="pb_copyResults">
```
dail8859, line 34: I know I'm being overly nit-picky but something like `btnCopyResults` would be more consistent with the rest of the UI elements in the other parts of the application.

dail8859 (changes requested):
Few minor changes for now. Haven't had a chance to compile/run/test these but it all seems relevant.

dail8859:
I'll be busy for the next week or so but will get back to this. Appreciate the PR!

src/NotepadNext/docks/SearchResultsDock.cpp
pydavem: I have been unable at this point to locate any information on line ends for the clipboard, and I don't have access to windows to test it, but it is a good question :)

pydavem: I haven't found any way to copy the tree contents without walking it, but as always, open to suggestions from more knowledgable people :)

src/NotepadNext/docks/SearchResultsDock.h
pydavem: Yes, sorry, QTCreator suggested it and I took the automatic route instead of thinking ;)

src/NotepadNext/docks/SearchResultsDock.ui
pydavem: You are quite right again, I should have looked ahead and followed the design of the existing code, not just throwing my own style in there. Attention to detail!!

src/NotepadNext/docks/SearchResultsDock.cpp
dail8859: I'll try it out on windows and see what happens :)

dail8859: I'm not sure either. Was just curious. I'll see if there is anything else obvious.

dail8859:
Visually the code changes look fine. Once I get a chance I'll compile the code and do some testing.

dail8859:
I pushed some changes. I'm sure there's room for improvement, especially when compared to Notepad++'s implementation...but perfection is the enemy of progress.

I redid the button layout to make it a bit more extensible in the future if needed.

Also redid how it built the string...I believe that will minimize the number of QString allocations on the heap.

Also using `\n` didn't seem to have a negative impact so I'll leave it for now. I usually wait until there is an actual problem before assuming something needs fixed.

There's also one other small quirk I found with it but doubt anyone will ever find it...and the fix isn't obvious...so I'll leave it alone :)

If all looks good to you I'll merge this in.

pydavem:
Nice changes, looking good thanks!

PR #623, opened 2024-08-15
Fix some crashes with Qt 6.7.2

sandman7920:


dail8859 (approved):
Awesome thanks! Just keep in mind there are some issues with Qt >6.5 with some of the 3rd party libraries, but definitely good to try to address any issues if/when Notepad Next is able to update the Qt version.

Issue #628, opened 2024-08-18
Backward direction not work in Find 


accessv:
### Operating System and Version

macOs

### Distribution

brew

### Description

With and without the selection of the "Backward direction" in the Find dialog window, the search will be only in the forward direction.

### Steps to Reproduce

1. open the Find window (command + F) 
2. set finding pattern
3. set check on Backward direction 
4. the search will be in a forward direction 

### Additional Details

e.g. screenshots, logs, etc

dail8859:
@accessv Could you open a PR with these changes? I'd be happy to review/test!

dail8859:
Closed by #635

PR #635, opened 2024-08-21
Add backward direction to Find dialog

accessv:
Fix for issue "Backward direction does not work in Find #628"


dail8859:
Thanks!

PR #652, opened 2024-10-26
i18n: Update all i18n .ts files

inkydragon:


dail8859:
Thanks!

I have a workflow https://github.com/dail8859/NotepadNext/blob/master/.github/workflows/lupdate.yml to run lupdate but have to trigger it manually. I should probably put it on a schedule so that the ts files stay updated.

PR #653, opened 2024-10-26
Update to Scintilla v5.5.3, Lexilla v5.4.1

inkydragon:


inkydragon:
Since Lexilla is move to GitHub https://github.com/ScintillaOrg/lexilla
Maybe we can add it to git submodule.

dail8859:
Thanks!

> Maybe we can add it to git submodule.

I've considered this. The _idea_ of submodules is good but every time I deal with them I end up regretting it. They end up just being a pain.

Issue #669, opened 2024-11-26
NotepadNext 0.9 does not build against upcoming GCC 15


SigHunter:
### Operating System and Version

Gentoo Linux

### Distribution

ebuild, compiled from source

### Description

I packaged notepadnext for Gentoo GURU Overlay and was made [aware](https://bugs.gentoo.org/945001), that it does not compile with upcoming GCC Version 15 and I am able to reproduce this aswell. The issue seems to be with Scintilla, I hope opening a bug report for NotepadNext is the correct way

GCC Version: 15 git Snapshot 2024-11-24
Qt Version: 6.7.3
Kernel Version: 6.11.10-gentoo-dist (64-bit)

`ScintillaTypes.h:693:18: error: intptr_t does not name a type`

Please find the build log attached
[build.log](https://github.com/user-attachments/files/17921074/build.log)

### Steps to Reproduce

1. Install GCC 15, for Gentoo this is sys-devel/gcc-15.0.0_pre20241124 and set current compiler profile with gcc-config 
2. Download https://github.com/dail8859/NotepadNext/archive/refs/tags/v0.9.tar.gz
3. tar xfv v0.9.tar.gz, cd NotepadNext-0.9, mkdir build, cd build, qmake6 ../src/NotepadNext.pro, make -j16

dail8859:
[Scintilla](https://scintilla.org) is a 3rd party open source library. So if there are any direct changes needed to Scintilla files then an issue would need opened with that project. They may need to make some changes for newer gcc versions.

That being said, Scintilla uses it's header files in an odd way, so not sure if this can be fixed on the "Notepad Next" side such as including additional header files before including the Scintilla files, etc. But I'm not sure since so far things working with current compilers. If you do find a fix that this project can do then I'm happy to include them assuming they are minimal and don't affect other build systems.

Also...I know this compiles with newer versions of Qt, but there's some app issues with anything >6.5.x so if you stick to Qt 6.7 your milage may vary.

SigHunter:
> [Scintilla](https://scintilla.org) is a 3rd party open source library. So if there are any direct changes needed to Scintilla files then an issue would need opened with that project. They may need to make some changes for newer gcc versions.
> 
[...]

PR #670, opened 2024-11-28
Include cstdin in scintilla for GCC15 compatibility

SigHunter:
I had to add cstdin in 3 places to make it compile successfully with the upcoming gcc15, because cstdin is not automatically included anymore.  Reference: https://gcc.gnu.org/gcc-15/porting_to.html and https://sourceforge.net/p/scintilla/bugs/2458/

I am no expert, so I don't know what style guidelines are custom for the placement of those includes.  If necessary, I can adapt the commits. Let me know what you think, thanks

The corresponding bug report is #669

### DIFF
--- a/src/scintilla/include/ScintillaTypes.h
+++ b/src/scintilla/include/ScintillaTypes.h
@@ -10,6 +10,7 @@
 
 #ifndef SCINTILLATYPES_H
 #define SCINTILLATYPES_H
+#include <cstdint>
 
 namespace Scintilla {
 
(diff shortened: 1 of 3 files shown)


dail8859:
Normally I leave any of the Scintilla stuff completely unmodified but willing to merge in these minor changes knowing that the upstream Scintilla project will include these soon.

I would also make sure the Scintilla bug report knows about the other place you had to include the header for the qt specific files.

dail8859:
Ah looks like they did already.

https://sourceforge.net/p/scintilla/code/ci/c7ffad21b23dfad4e8e9f36bb587acc2f6e84fee/

PR #721, opened 2025-02-19
Update Spanish (Nicaragua, es_NI) translation.

Ursamenux:
Here's my second pull request to the `master` branch, regarding the Nicaraguan Spanish translation with the following changes:

1. Fix some minor spelling mistakes I did during the translation process.
2. Translate some strings that had no Spanish translation.
3. Fix some Spanish translation strings so they can (kind of) make sense for Spanish speakers.


dail8859:
Thanks!

Issue #779, opened 2025-05-27
Error while building member reference base type 'int' is not a structure or union


matthewyang204:
### Operating System and Version

macOS 15 Sequioa

### Distribution

src

### Description

When built from /build via `qmake ../src/NotepadNext.pro && make -j16`, it gives this error while building:
```
../../src/NotepadNext/dialogs/MainWindow.cpp:868:52: error: member reference base type 'int' is not a structure or union
  868 |             if (sequence.count() > 0 && sequence[0].key() != Qt::Key_unknown) {
      |                                         ~~~~~~~~~~~^~~~
```

### Steps to Reproduce

1. Make a directory /build inside of the cloned repo
2. `qmake ../src/NotepadNext.pro`
3. `make -j16`

### Debug Information

N/A because the app refuses to build, let alone use it

### Additional Details

```
../../src/NotepadNext/dialogs/MainWindow.cpp:868:52: error: member reference base type 'int' is not a structure or union
  868 |             if (sequence.count() > 0 && sequence[0].key() != Qt::Key_unknown) {
      |                                         ~~~~~~~~~~~^~~~
```

dail8859:
What version of Qt are you using?

matthewyang204:
I'm using Qt5.15, sorry, I forgot to look. I'll try again with Qt6

dail8859:
I have no clue when it comes to building on mac os. After doing some reading/research/chatgpt it might help to take a look here: https://doc.qt.io/qt-6/macos.html#architectures

It looks like you might want to try adding `QMAKE_APPLE_DEVICE_ARCHS = x86_64 arm64` in the qmake file.

dail8859:
Glad it is working now. In theory it should be easy to build the project with Qt5.15 with minor code changes. I don't plan on fixing any Qt5 specific issues, however anyone wanting to submit fixes for older Qt versions will be accepted as long as they do not introduce complex changes.

matthewyang204:
Yeah, if it's possible to do function based on the version of qt you're running, and the two alternate implementations are drop-in replacements for each other, should be easy.

PR #782, opened 2025-05-31
Fix Qt 5 compatibility and add Find All in Files button

matthewyang204:
This fixes issue #779 on Qt5 while retaining Qt6 compatibility. It now includes a separate block using the #if and #elif blocks to determine which block runs depending on Qt version. It also adds a Find All in Files button. Looks like most of what is needed to implement Find in Files is ready, so someone else can probably do the actual functionality, and therefore, I'll include this button in the Find and Replace dialog. The .gitignore is also updated to ignore cache files created by some popular C++ IDEs and all build folders (some people like me like to create numbered build folders to preserve older builds).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,11 @@ CMakeLists.txt.user*
 /build
 /icon/*.png
 /icon/ImageMagick
+
+# JetBrains IDE prefs
+.idea/
+
+# macOS annoyances
+.DS_Store
+*DS_Store
+._*
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ CMakeLists.txt.user*
 *.pyc
 
 /build
+/build*
 /icon/*.png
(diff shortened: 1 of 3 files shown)


matthewyang204:
Yep, works for me. Just please add the button sometime after someone adds a proper Find in Files function. I do remember Notepad++ has a Find All and Replace All in the dialog for files in the folder opened as workspace.

matthewyang204:
Alright, I just tested, build succeeded in Xcode, so it all works fine

matthewyang204:
Just tested on Debian 12 (Linux 6.1.x) with Qt 5 on my Linux rig. Build completed successfully. Note that this rig cannot run Qt 6, as it is running Debian Stable (Bookworm 12).

matthewyang204:
Alright, resetting my fork to your repo and then testing it, I built it again. Build succeeded. All is good

Issue #780, opened 2025-05-27
Editor Context Menu Location


dail8859:
### Description

When right-clicking in the editor there is inconsistent behavior because some commands use the position of the mouse to determine if some things can be used (e.g. a URL), but others need the current caret position (markers).

### Describe the solution you'd like

It seems most editors rely on the logic:

- If the right click is inside a selection, do not change the selection at all, then open the context menu based on the current selection.
- If the right click is outside the selection, place the caret as the mouse position, then open the context menu based on the new location.

### Describe alternatives you've considered

None - Most editors seem to have the described behavior.

migsan1104:
Hello, I am requesting to merge my fix to this issue. When you get a chance can you pls take a look at it?

PR #783, opened 2025-06-05
Fix:#780 Miguel/editor fix

migsan1104:
This PR addresses issue [#780](https://github.com/dail8859/NotepadNext/issues/780) by implementing behavior consistent with modern editors:

If the user right-clicks inside an existing selection, the selection is preserved.

If the user right-clicks outside the selection, the caret moves to the clicked position and the context menu is invoked at the new location. The selection is also cleared.

Added build/ to .gitignore to avoid tracking generated files.


migsan1104:
Thank you for letting me work on this project. :)

dail8859:
Hi @migsan1104 apologies for the delay.

Although I think this is likely a valid solution, all the code within the `scintilla` directory is directly from a 3rd party library so I wouldn't want to maintain a separate copy or need to remember to edit it every time.

My first guess is that this section of code is the best place to handle it:

https://github.com/dail8859/NotepadNext/blob/6ab46c42b7c3b5887e757a5ff4cb04398a76f818/src/NotepadNext/dialogs/MainWindow.cpp#L1973-L2010

That is where the main window shows the context menu and may need to do some checking to see if its in a selection or not and possibly clear the selection.

Notepad++ also has this behavior and also uses Scintilla so it might have a recommended way of handling this case, but I havent taken the time to dig into the Notepad++ source code in quite a while.

PR #796, opened 2025-06-18
Update-deps (git submodules)

dafanasiev:
subj


dail8859:
Just curious, does this fix any particular issue you are having? Or is this just to update the submodules just for the sake of updating?

dafanasiev:
> Just curious, does this fix any particular issue you are having?

No, I think it's good practice to update dependencies when they are released new "stable" versions.

dail8859:
Looking at the actual changes there is really no significant code changes. Maybe the uchardet library but that library is barely even used.

The ADS library has not released another stable version. Even with that the last 'stable' version it was on had issues. I normally dont update submodules unless there's a good reason but I guess I see no harm in it.

Issue #827, opened 2025-07-06
Search Results fonts


dail8859:
### Discussed in https://github.com/dail8859/NotepadNext/discussions/823

<div type='discussions-op-text'>

<sup>Originally posted by **rayray519** July  5, 2025</sup>
Noticed the search results fonts are not the same as the main window.  Is there plans to allow to resize the fonts or have them the same (or behave the same as Notepad ++)? </div>

dail8859:
If someone is interested in fixing this, this section:

https://github.com/dail8859/NotepadNext/blob/6f27d7df09063d9c3d2bf302054b4a10dfcac3f4/src/NotepadNext/docks/SearchResultsDock.cpp#L39-L43

can be updated. Just use the application settings font and size and not do anything specific per OS.

kolcz:
@dail8859 I could take it. Is it OK?

dail8859:
Sure! 😊

PR #834, opened 2025-07-12
Unify Font Size In Search Result Dock

kolcz:
## Description
This PR addresses issue #827. In this PR I made font, in SearchResultsDock to match application settings parameters during initialization and when font is changed in PreferencesDialog.

## Changes Made
- Remove setting different font size on MacOS,
- Add setFont() method to SearchResultsDock,
- Added running setFont after SearchResultsDock initialization,
- Connect ApplicationSettings::fontNameChanged and ApplicationSettings::fontSizeChanged to setFont method.

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing - on Windows 11 and Debian 12 bookworm
- [x] Font in SearchResultsDock, Editor are the same during initialization and identical to font specified in PreferencesDialog,
- [x] After changing font size by clicking arrows, both fonts in Editor and SearchResultsDock are modified,
- [x] After changing font size by typing new value, both fonts in Editor and SearchResultsDock are modified,
- [x] After  changing font family by selecting from the list, both fonts in Editor and SearchResultsDock are changed,
- [x] After changing font family by typing new value, both fonts in Editor and SearchResultsDock are changed.

## Additional Note
Someone need to test it on MacOS.

### DIFF
--- a/src/NotepadNext/dialogs/MainWindow.cpp
+++ b/src/NotepadNext/dialogs/MainWindow.cpp
@@ -317,9 +317,6 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
     srDock->toggleViewAction()->setShortcut(Qt::Key_F7);
     ui->menuView->addAction(srDock->toggleViewAction());
 
-    ApplicationSettings *settings = app->getSettings();
-    srDock->setFont(settings);
-
     connect(srDock, &SearchResultsDock::searchResultActivated, this, [=](ScintillaNext *editor, int lineNumber, int startPositionFromBeginning, int endPositionFromBeginning) {
         dockedEditor->switchToEditor(editor);
 
@@ -329,9 +326,6 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
 
         editor->grabFocus();
     });
-
-    connect(settings, &ApplicationSettings::fontNameChanged, srDock, [=]() { srDock->setFont(settings); });
-    connect(settings, &ApplicationSettings::fontSizeChanged, srDock, [=]() { srDock->setFont(settings); });
 
     connect(ui->actionFind, &QAction::triggered, this, [=]() {
         showFindReplaceDialog(FindReplaceDialog::FIND_TAB);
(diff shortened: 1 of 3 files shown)


dail8859:
Thanks for this!

I think this is a good first step. It would be best to also go ahead and always set the font of the tree view to the default font and size that is specified in the ApplicationSettings. This can be changed in the future but it's a logical step so that things will be consistent.

kolcz:
You're welcome. :)
Yeah, sure. Good idea! I was a little surprised when I found out that Debian by default has font "Liberation Mono" with size 12, so it will be a good idea to unify that.

kolcz:
@dail8859 Could you take a look at changes? I can't request a review by github now. :/

dail8859:
Apologies for the long delay.

I consolidated the changes into just the SearchResultsDock. It also got rid of overloading the `setFont()` method already defined by `QWidget`. This way the MainWindow doesn't need do anything else specifically for the dock. It might be a bit terse for now but it seems to work correctly.

Let me know if you have any thoughts.

kolcz (commented):
I think it's better to implement all logic inside `SearchResultsDock`. It will isolate particular docks windows and when we implement a lambda function like we don't pass variables from another objects. My problem was that  I didn't know how to get `ApplicationSettings` from inside `SearchResultsDock` class.
I wonder what does `resizeColumnToContents` do? As I tested when font increase the text can move out of window space. So I don't see how this function work.
I tested both on Windows 11 and Debian and it works as it work as intended. I think we can merge Pull Request. Or you see another way to improve it?

dail8859:
Technically you can instantiate a `ApplicationSettings` anywhere you like. The instantiation is actually quite light weight since it is based on `QSettings`. However, to properly trigger and also receive notifications you need to make sure to get to the pointer to the one that is owned by the `NotepadNextApplication` instance.

> I wonder what does resizeColumnToContents do?

Without this line, the column used for the line numbers does not dynamically resize leading to the line numbers getting : 
resize

kolcz:
Thank you for the explanation and help with new feature. Now I better understand the application.

PR #852, opened 2025-08-17
Fix multi-instance file path passing

coladict:
I tried opening a file through the console while I had an instance running, but the primary instance looked for it in its own working directory.

Not sure if this code is up to standard, but it works. I haven't written in C++ in about 13 years.

### DIFF
--- a/src/NotepadNext/NotepadNextApplication.cpp
+++ b/src/NotepadNext/NotepadNextApplication.cpp
@@ -341,8 +341,19 @@ void NotepadNextApplication::sendInfoToPrimaryInstance()
 
     QByteArray buffer;
     QDataStream stream(&buffer, QIODevice::WriteOnly);
+    QStringList argsToSend;
+    QStringList toFilter = parser.positionalArguments();
+    for (const QString &arg : arguments()) {
+        if (toFilter.contains(arg)) {
+            QFileInfo fileInfo(arg);
+            argsToSend.push_back(fileInfo.absoluteFilePath());
+        }
+        else {
+            argsToSend.push_back(arg);
+        }
+    }
 
-    stream << arguments();
+    stream << argsToSend;
     const bool success = sendMessage(buffer);
 
     if (!success) {


dail8859:
Thanks for the PR!

Could you provide a set of steps of the issue you are having? That way I can verify the code changes and try to reproduce it on Windows.

dail8859:
Also on the primary instance if you open the `Help > Debug Log` it will show what it is getting sent which might be helpful to see what is getting sent across.

coladict:
Steps to reproduce issue:

1. Open NotepadNext from launcher (runs with home directory as work directory)
2. In any terminal go to directory other than home.
3. Attempt to open file from that directory, eg `NotepadNext main.cpp`
4. It will open "main.cpp" in the home directory, instead of the current one for the original invocation.

dail8859:
👍

coladict:
Log from the affected version:
```
[     8.560] I: void NotepadNextApplication::receiveInfoFromSecondaryInstance(quint32, QByteArray)
[     8.561] I: void NotepadNextApplication::openFiles(const QStringList&)
[     8.561] I: void MainWindow::openFileList(const QStringList&)
[     8.561] I: 01.sql
[     8.561] D: Read 0 bytes
```

Log from this version (with the home name censored):
```
[     5.677] I: void NotepadNextApplication::receiveInfoFromSecondaryInstance(quint32, QByteArray)
[     5.677] I: void NotepadNextApplication::openFiles(const QStringList&)
[     5.677] I: void MainWindow::openFileList(const QStringList&)
[     5.677] I: /home/user/db-update-scripts/3.12.0/01.sql
[     5.678] D: Read 248 bytes
```

coladict:
Frankly the change to use a Set instead of the list was unnecessary. If you ever have a command line with enough files that you can measure a performance improvement, you're using it wrong.

dail8859:
Honestly you are completely right. I was testing out some other things with that at the time and tried to wrap it up late at night and definitely the way it is implemented is not ideal for the 99.99% case. I tend to not trust user input but as a certain point you just tell the user "don't do that".

dail8859:
https://github.com/dail8859/NotepadNext/commit/993120a19c12955f46626ff6d1bfb2fe1b2ca8bd

Issue #845, opened 2025-08-06
Corrupt session causes application to not load


ck62:
### Operating System and Version

macOS Sequoia Version 15.6

### Distribution

Homebrew

### Description

After updating to macOS Sequoia Version 15.6, NotepadNext 0.12.0 no longer runs on the Mac.

Error report for Notepad Next
Notepad Next quit unexpectedly.
This report is automatically sent to Apple.

Problem details and system configuration
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               NotepadNext [19962]
Path:                  /Applications/Notepadnext.app/Contents/MacOS/NotepadNext
Identifier:            io.github.dail8859.NotepadNext
Version:               0.12.0 (???)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2025-08-06 11:38:11.8728 +0200
OS Version:            macOS 15.6 (24G84)
Report Version:        12
Anonymous UUID:        B5749303-626D-59A8-8E42-292EDD8814C7

Time Awake Since Boot: 3000 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000010
Exception Codes:       0x0000000000000001, 0x0000000000000010

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [19962]

VM Region Info: 0x10 is not in any region.  Bytes before following region: 4362485744
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                      104064000-10450c000    [ 4768K] r-x/r-x SM=COW  /Applications/Notepadnext.app/Contents/MacOS/NotepadNext

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   NotepadNext                   	       0x1042c8978 BookMarkDecorator::setBookMarkedLines(QList<int>&) + 24
1   NotepadNext                   	       0x1042c0390 SessionManager::loadEditorViewDetails(ScintillaNext*, QSettings&) + 528
2   NotepadNext                   	       0x1042bf6d8 SessionManager::loadUnsavedFileDetails(QSettings&) + 580
[...]

Issue #658, opened 2024-11-12
Session restoration only works upon close


matthewhively:
### Description

The automatic session save for open tabs apparently only works when the window is closed.
In notepad++ the session is saved periodically so that if there is an unexpected error/crash/reboot (whatever) your session has a reasonable chance to remain intact.

### Describe the solution you'd like

I would prefer if session was backed up at least every hour if there were any changes made to open documents, if not more often.

PR #874, opened 2025-10-03
Autosave Session (60 seconds timer)

### LINKED ISSUE #871: changes are saved only on a clean exit
### Description

I noticed that if I change buffer but don't save file and don't do clean exit my changes are not flushed to the disk. As a result in case of an OS crash my changes would be lost. 

It would be really nice if notepadnext will run scheduled checks if buffers are updated (e.g. every minute) and if there were changes it will flush them to disk.

_No response_

SimoSbara:
This is a simple implementation of a QTimer that saves the sessione each 60 seconds.
It could be improved by adding the interval inside the Preferences Dialog.

It fixes #871 and #658

### DIFF
--- a/src/NotepadNext/NotepadNextApplication.cpp
+++ b/src/NotepadNext/NotepadNextApplication.cpp
@@ -440,6 +440,17 @@ void NotepadNextApplication::saveSettings()
     getSettings()->setValue("App/RecentFilesList", recentFilesListManager->fileList());
 }
 
+void NotepadNextApplication::saveSession()
+{
+    for (const auto &editor : window->editors()) {
+        if (editor->isFile()) {
+            recentFilesListManager->addFile(editor->getFilePath());
+        }
+    }
+
+    getSessionManager()->saveSession(window);
+}
+
 MainWindow *NotepadNextApplication::createNewWindow()
 {
     Q_ASSERT(window == Q_NULLPTR);
@@ -452,15 +463,12 @@ MainWindow *NotepadNextApplication::createNewWindow()
     });
 
     // Since these editors don't actually get "closed" go ahead and add them to the recent file list
-    connect(window, &MainWindow::aboutToClose, this, [=]() {
-        for (const auto &editor : window->editors()) {
-            if (editor->isFile()) {
-                recentFilesListManager->addFile(editor->getFilePath());
-            }
-        }
+    connect(window, &MainWindow::aboutToClose, this, &NotepadNextApplication::saveSession);
 
-        getSessionManager()->saveSession(window);
-    });
+    // Timer to autosave session each minute
+    connect(&autoSaveTimer, &QTimer::timeout, this, &NotepadNextApplication::saveSession);
+    autoSaveTimer.setInterval(60 * 1000);
+    autoSaveTimer.start();
 
     return window;
 }
--- a/src/NotepadNext/NotepadNextApplication.h
+++ b/src/NotepadNext/NotepadNextApplication.h
@@ -26,6 +26,7 @@
 
 #include <QCommandLineParser>
 #include <QPointer>
+#include <QTimer>
 
 
 class MainWindow;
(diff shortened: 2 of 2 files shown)


SimoSbara:
The latest commit could fix #845

src/NotepadNext/SessionManager.cpp
```diff
@@ -316,10 +316,10 @@ ScintillaNext *SessionManager::loadUnsavedFileDetails(QSettings &settings)
         editor->setFileInfo(filePath);
         editor->setTemporary(true);
 
-        loadEditorViewDetails(editor, settings);
-
         app->getEditorManager()->manageEditor(editor);
 
+        loadEditorViewDetails(editor, settings);
```
dail8859, line 321: was there a specific reason this was moved after `app->getEditorManager()->manageEditor(editor);`?

src/NotepadNext/NotepadNextApplication.cpp
```diff
@@ -440,6 +440,17 @@ void NotepadNextApplication::saveSettings()
     getSettings()->setValue("App/RecentFilesList", recentFilesListManager->fileList());
 }
 
+void NotepadNextApplication::saveSession()
+{
+    for (const auto &editor : window->editors()) {
+        if (editor->isFile()) {
+            recentFilesListManager->addFile(editor->getFilePath());
+        }
+    }
```
dail8859, line 449: The critical part is definitely saving the session. If there was a crash I dont think the recent file list is that big of deal. Do you think its still a good idea to leave in the `saveSession()` or only just call it during `aboutToClose()`. I guess I don't have a preference, this was just a thought.

SimoSbara: I would save everything even if it is not critical. Also, in the future where there will be new features, `saveSession()` will still be useful.

src/NotepadNext/SessionManager.cpp
SimoSbara: The inversion resolved a crash, otherwise inside `loadEditorViewDetails(editor, settings)` the decorator reference is null because is not created yet! `app->getEditorManager()->manageEditor(editor);` does the allocations.

src/NotepadNext/NotepadNextApplication.cpp
dail8859: I think given the current design I think what you have definitely makes the most sense for now. There's still lots of things I would like to clean up, refactor, and rearchitect.

dail8859:
The good thing about QTimer is it will run the session saving on the main thread. Scintilla is not thread safe. However, that being said it could possibly cause temporary lag spikes backing up the session. Have you tested this at all on having multiple and/or large files opened? I'm curious if there's any indication of when it may start to visually lag the UI and input for the user.

In the end this is by far the safest/easiest.

SimoSbara:
> The good thing about QTimer is it will run the session saving on the main thread. Scintilla is not thread safe. However, that being said it could possibly cause temporary lag spikes backing up the session. Have you tested this at all on having multiple and/or large files opened? I'm curious if there's any indication of when it may start to visually lag the UI and input for the user.
> 
> In the end this is by far the safest/easiest.

I often use QTimer for doing things (qt graphics, signals to UI in qt quick) from the main thread. 
About the lag spikes: yes, it could happen but I think it is unnoticeable in the normal use case. I will test it on a few large files and let you know if it lags a lot. 

Another idea could be:
1. the main thread retrieves the text buffer from Scintilla
2. the main thread passes the text buffer to another thread that writes on the disk, this avoids the UI to lock for the disk write operation, the main threads needs to pass the memory reference to another thread (very fast).
3. the main thread continues the execution of UI

This happens only in the "autosave" done by the QTimer. The `aboutToClose()` could call a slot that saves everything from the main thread.

dail8859:
> it could happen but I think it is unnoticeable in the normal use case.

Agreed. 

> Another idea could be:

Ah yes that would prevent accessing the Scintilla editor(s) from multiple threads by grabbing and passing the _text_ to the background thread, rather than the thread using the _editor_. It could possibly have a memory spike "duplicating" all the files, but that is probably also acceptable. Or it could pass the text buffer one by one rather then all of them at once. Just an implementation difference is all.

---

Overall I think this is a great and simple first implementation to see how well it works in the real world. No point in over-engineering a solution if it's not need yet.

PR #880, opened 2025-10-17

matthewyang204:
Changes:
1. Rename the Find in Files button to Replace in Files because Find in Files is 1. pretty useless 2. too hard to implement (unless someone else wants to do so)
2. Build a Replace in Files where the user can replace stuff in a directory of their choice recursively

Please check that you like it, and then make any changes you want. Otherwise, feel free to merge. Also, given the number of commits, you may not want to squash the entire thing into one patch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,28 +104,28 @@ jobs:
           make appimage
 
       - name: Upload Windows zip Package
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         if: runner.os == 'Windows'
         with:
           name: NotepadNext-Windows-Qt${{ matrix.qt_version }}-Zip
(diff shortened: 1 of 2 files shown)


matthewyang204:
Build succeeded on Qt 6 on Ubuntu 24.04 LTS & Qt 5 on Debian 13 Trixie

matthewyang204:
Also, to note, I also get some occasional UI glitches on my Debian tower under GNOME

matthewyang204:
Alright, the UI is less screwed up after I added those two commits to make it so the window can't be shrunk to zero size.

matthewyang204:
Also, to note, this only works on ASCII & UTF-8 files. However, it is reliable enough and besides, this editor doesn't even have support for other encodings yet anyways

matthewyang204:
@dail8859 Can you please take a look at this PR? It's been a long time.

PR #899, opened 2025-12-05
Bump Ubuntu runner to 24.04 LTS

matthewyang204:
This bumps the Ubuntu version used for the Linux runner to 24.04 LTS because that is the newest LTS version available on GitHub runners.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,28 +104,28 @@ jobs:
           make appimage
 
       - name: Upload Windows zip Package
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         if: runner.os == 'Windows'
         with:
           name: NotepadNext-Windows-Qt${{ matrix.qt_version }}-Zip
           path: ${{ github.workspace }}/build/package/
 
       - name: Upload Windows Installer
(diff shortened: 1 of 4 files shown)


dail8859:
The only reason I might be cautious to merge this is for some issues experienced in the past:

https://github.com/dail8859/NotepadNext/issues/77 fixed by https://github.com/dail8859/NotepadNext/pull/78

To support a wider/older varieties of Linux it was changed to keep an older version of Linux.

I'm not a Linux user so I'm not sure what all compatibilities this may introduce. If its not foreseen to cause issues I'm fine updating the runner.

matthewyang204:
Oh okay I understand - we need @pgarciaq to test the build artifacts of this PR before we can merge. Then we can decide.

pgarciaq:
@matthewyang204 I no longer have a RHEL8 box to test. Those builds work fine on Fedora 43, though.

matthewyang204:
@pgarciaq I'm pretty sure it'll be fine then. RHEL8 was released years ago and most users won't be using it anymore. And besides, the beauty of Linux is you can build from src anytime you want. And most Ubuntu users are now on 24.04 LTS Noble (including me), not Jammy or Focal. I suppose it's safe to say it'll be alright.

PR #906, opened 2025-12-16

matthewyang204:
Changes:
1. Rename the Find in Files button to Replace in Files because Find in Files is 1. pretty useless 2. too hard to implement (unless someone else wants to do so)
2. Build a Replace in Files where the user can replace stuff in a directory of their choice recursively

Please check that you like it, and then make any changes you want. Otherwise, feel free to merge. Also, given the number of commits, you may not want to squash the entire thing into one patch

### DIFF
--- a/i18n/NotepadNext_it.ts
+++ b/i18n/NotepadNext_it.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.0" language="it">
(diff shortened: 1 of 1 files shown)


matthewyang204:
@dail8859 Can you please check out this PR? It's been a long time and you have not checked the previous PR I opened, so I re-opened this one with the same commits.

matthewyang204:
I broke this PR when I renamed the branch as I was sorting out my fork. Please go to #913, which is the reopened PR.

PR #913, opened 2026-01-11

matthewyang204:
> Changes:
> 1. Rename the Find in Files button to Replace in Files because Find in Files is 1. pretty useless 2. too hard to implement (unless someone else wants to do so)
> 2. Build a Replace in Files where the user can replace stuff in a directory of their choice recursively
> 
> Please check that you like it, and then make any changes you want. Otherwise, feel free to merge. Also, given the number of commits, you may not want to squash the entire thing into one patch

Sorry, I renamed the branch which caused #906 to close. I've reopened it here.

### DIFF
--- a/i18n/NotepadNext_it.ts
+++ b/i18n/NotepadNext_it.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.0" language="it">
+<TS version="2.1" language="it">
 <context>
     <name>AuthenticateDialog</name>
     <message>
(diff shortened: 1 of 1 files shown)


dail8859:
My apologies. I was reviewing this and did not mean to merge it.

The part that I do not like about this PR is that it does not use any of the underlying functionality (regex...which I know does not work great, extended mode, match case, whole word only, etc) which will cause confusion to users expecting it to work exactly the same as the normal replace functionality to files that are opened.

matthewyang204:
We can always fix it though, thanks for feedback!

PR #920, opened 2026-01-28
Fix OOB read vulnerability in kname() in lua/ldebug.c

biniamf:
Fixes an potential out-of-bounds read in kname() by validating bounds.

This prevents integer overflow vulnerability potentially leading to information leak or arbitrary code execution when loading malicious binary chunks


dail8859:
Thanks!

Issue #928, opened 2026-01-31
Tarball release


drfiemost:
### Description

The github's automatically generated source package lacks the submodules and is basically useless.
It would be nice to have a complete tarball release available for those who build from sources.

### Describe the solution you'd like

I've used the command described [here](https://stackoverflow.com/a/67389978) to generate the tarball locally

`git ls-files --recurse-submodules | tar caf ../NotepadNext-0.13.tar.xz --xform s:^:NotepadNext-0.13/: --verbatim-files-from -T-`

matthewyang204:
Yeah would be nice. I do also prefer a tarball release that's manually generated whenever there's submodules.

dail8859:
Unfortunately it's nothing I can opt in/out of or control through GitHub. I'm sure it's come up before with other repositories and ultimately it would be nice if GitHub natively supported it.

If someone wanted to submit a PR to update the workflow to publish a full source code release I'd be willing to accept it.

drfiemost:
> If someone wanted to submit a PR to update the workflow to publish a full source code release I'd be willing to accept it.

Done in #932

PR #932, opened 2026-02-01
Create tarball release workflow (#928)

drfiemost:


dail8859:
Hi @drfiemost Thanks for this PR.

Would it be possible to integrate the key steps in that workflow file into the existing `build.yml`. This is the section that runs when there is a tag and creates a draft release.

https://github.com/dail8859/NotepadNext/blob/8ba0ed89fcc685b541b4418690c28d2804b4fd74/.github/workflows/build.yml#L135-L174

That way it stays in the same place as the other artifacts that are uploaded to a draft release, and also uses `gh release upload ...` and isn't dependent on another action e.g. `JasonEtco/upload-to-release@master`

---

I think it would be something like...set the step `Checkout Repository` to recursive, and then add these two sections (completely untested)

```yaml
- name: Create source tarball
  run: |
        export PACKAGE_NAME=NotepadNext-${{ github.ref_name }}
        export RELEASE_TARBALL=${PACKAGE_NAME}.tar.gz
        git ls-files --recurse-submodules | tar caf ${RELEASE_TARBALL} --xform s:^:${PACKAGE_NAME}/: --verbatim-files-from -T-

- name: Upload source tarball
  run: |
    gh release upload ${{ github.ref_name }} NotepadNext-${{ github.ref_name }}.tar.gz
```

matthewyang204:
Please `git am` these two patches and test; I've moved the stuff to build.yml:
[0001-Move-create_tarball.yml-s-stuff-to-build.yml.patch](https://github.com/user-attachments/files/25049890/0001-Move-create_tarball.yml-s-stuff-to-build.yml.patch)
[0002-Remove-whitespace.patch](https://github.com/user-attachments/files/25049891/0002-Remove-whitespace.patch)

matthewyang204:
Please also `git am` this patch to your PR branch. We currently have two "upload" steps for the source tarball which will cause conflicts, as the first one even runs before the creation step. This patch removes the duplicate first one.
[0001-Remove-extra-Upload-Source-Tarball-step.patch](https://github.com/user-attachments/files/25059375/0001-Remove-extra-Upload-Source-Tarball-step.patch)

drfiemost:
Applied.
Unfortunately I'm unable to test this on my repo, it always skips the GitHub Release step

matthewyang204:
Alright just going to test on my own fork then

matthewyang204:
@drfiemost It does work, and the src ZIP contains no binaries. It works very well

dail8859:
Appreciate the PR! Looks good to me!

PR #944, opened 2026-02-09
fix: Support building in QT5, QByteArrayView compat

savchenko:
As of https://github.com/dail8859/NotepadNext/commit/637480688c7d879fb0e92d571717416089c315ab, the build will fail with `QByteArrayView: No such file or directory`:

```
In file included from ../../src/NotepadNext/ScintillaNext.cpp:24:
../../src/NotepadNext/ByteArrayUtils.h:21:10: fatal error: QByteArrayView: No such file or directory
   21 | #include <QByteArrayView>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile:7065: build/obj/ScintillaNext.o] Error 1
make[1]: *** Waiting for unfinished jobs....
In file included from ../../src/NotepadNext/NotepadNextApplication.h:23,
                 from ../../src/NotepadNext/SessionManager.cpp:25:
../../src/NotepadNext/ApplicationSettings.h:47:12: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
   47 |     inline const char * const key() const { return mKey; }
      |            ^~~~~
```

This makes it buildable on Debian 13 using the set of packages supplied in `Building.md`. 
`QByteArrayView` referenced in `ByteArrayUtils.h` was introduced in QT6.

### DIFF
--- a/src/NotepadNext/ByteArrayUtils.h
+++ b/src/NotepadNext/ByteArrayUtils.h
@@ -18,9 +18,36 @@
 #pragma once
 
 #include <QByteArray>
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
 #include <QByteArrayView>
+#else
+#include <QtGlobal>
+#include <cstring>
+
+// WARNING: Dirty QT5 compat. Tested on Debian 13.
+class QByteArrayView
+{
+public:
+    QByteArrayView() : m_data(nullptr), m_size(0) {}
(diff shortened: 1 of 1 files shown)


dail8859:
This is not actively compiled with Qt5 which is why this broke, but appreciate the PR. I'm certainly willing to keep Qt5 compatibility as a best effort, and as long as it does not become burdensome...and someone else contributes to the fixes :)

This definitely seems like a reasonable compatibility shim to keep Qt5 working.

PR #945, opened 2026-02-09
Remove redundant const qualifier

savchenko:
`ApplicationSetting::key()`, silences `-Wignored-qualifiers`


dail8859:
That's alot of `const`ing :)

Thanks!

PR #946, opened 2026-02-09
feat: Harden build flags on *nix

savchenko:
Just following Debian best practises. Full RELRO with:
`-fstack-protector-strong`
`-D_FORTIFY_SOURCE=2`


dail8859:
Seems reasonable. Thanks!

Issue #937, opened 2026-02-02
Setup crowdin for translations


QuentiumYT:
### Description

Hello, I'd like to contribute for french translations but handling .ts files directly is quite boring ^^ I've use Crowdin for a Qt project I'm maintaining. (https://crowdin.com/project/stacer)
It's a simple setup (may take some time to fit every need) but works really well (and it's free)! If you would like some help, we can stay in touch.

### Describe the solution you'd like

My translation update process is:
1. Someone propose strings in Crowdin
2. I can review translations (before a commit is made) or not (I prefer not because it's quicker)
3. It creates 1 commit per translation file (in a period span of 1h I think) on a separate main_i10n branch
4. It creates also a new PR where I run a simple lupdate (for formatting style or update lines of the project)
5. I keep Crowdin PR open while there isn't much translations or merge it before a new release
6. Easy squash merge at the end :)

### Describe alternatives you've considered

Manual is very random / not complete / no review
Some other translation platforms I don't know about (self hosted exists)

skironDotNet:
Can't we just pull them from the original n++?

QuentiumYT:
Maybe that can work if the strings are the same, I am not sure if they are the same format. Also people translate them manually here as I can see

dail8859:
> Can't we just pull them from the original n++?

Not without a great deal of manual effort unfortunately.

@QuentiumYT Appreciate the suggestion. I don't ever deal with translating things myself so I'm open to ideas and glad to hear you've had success with it. I spent a few minutes getting logged in and seeing how to set it up. Also took a look at your stacer project as well. Seems to make sense but will try to find some more time soon to test out the integration more and will let you know if I have any questions on how you set yours up.

QuentiumYT:
Glad to hear that :) I know it can take some time to set up but people can translate it more easily. Don't hesitate to ping me if you have any questions!

pozemka:
I find that quite a lot of opensource projects are using Weblate. They have free plan for open source project. https://hosted.weblate.org/
I am personally contributed to project with both Crowdin and Weblate and find latter more friendly. But it is just personal subjective preference.

skironDotNet:
I code in C# not C++ so can't help much (unless used AI to refactor it), but if I had C++ I would work on this project full time, because I'm on Linux and would love to have native n++. Currently running Notepad++ via Wine and it works but loads slowly. Sorry this was off top.

[...]

PR #958, opened 2026-02-20
Cmake

dail8859:
I have no idea what I'm doing 🙃 

Overall:
- Moved all dependencies to the `thirdparty` folder.
- Removed all the git submodules (submodules are evil)
- Used `CPMAddPackage()` where feasible (could probably leverage it more but this is acceptable for now)
- All the code in `src/NotepadNext/*` was moved up one level since `src` is just the main app code
- This has not been tested on Linux/MacOS at all...I'm sure the build actions will fail
- Downstream projects will need to switch to cmake

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -30,7 +30,7 @@ jobs:
         run: |
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
-          git add src/Version.pri deploy/linux/com.github.dail8859.NotepadNext.metainfo.xml
+          git add CMakeLists.txt deploy/linux/com.github.dail8859.NotepadNext.metainfo.xml
           git commit -m "v${{ github.event.inputs.version }} release"
 
       - name: Tag commit


dail8859:
@diegoiast

I know you opened a PR quite a while ago for cmake and I learned alot reviewing that and also using ChatGPT :) Would love any insight you might have on this PR. Not looking for perfection but would like more of a 'sanity check' to know if there's anything obvious I'm overlooking. There are still improvements that can be made later for sure. 

@zocker-160

This PR will also mean updating the FlatHub build commands here: https://github.com/flathub/com.github.dail8859.NotepadNext/blob/cfc8b07a9974b78b8163d7cb8d5452d61b06ad23/com.github.dail8859.NotepadNext.yml

--- 

Of course anyone else is welcome to comment in this PR especially when it comes to Linux and MacOS because about the only thing I can verify there is that it builds 🤷

diegoiast:
Looked at it briefly, too large to view on github's UI. So random notes:

1. Moving the code a dir up - thanks. It hurted me back when looking at the code.
2. I downloaded the build on Linux, the AppImage worked. I do see an increase of 4MB between v0.12 and v0.13. You might want to look at it.
3. I don't see `lupdate` being run. Does CMake do it for you?
4. I see you opted to bring all the code into your repo, instead of using `FetchContent` as I was aiming to. Why was that?
5. What made be not finish the port back then, was the way qmake handled the `version.rc` or similar parts of the installer. I was hoping no to go the route you went - by using a post-processed file :)
6. I think you can remove `CPM.cmake` ... you are not using it right?
7. `commit_message: Update translation files  file_pattern: 'i18n/NotepadNext_en.ts'` I think you nuked translations. Are sure this is correct? (I see translations are available in the app, weird. Definitely something I do not understand).
8. You might want to use a different naming scheme for non official builds (using the git sha1 instead of 0.13). See here for reference: https://github.com/codepointerapp/codepointer/blob/a0a0cdeb767ef8310c4ee8a7fcaa7eadbf666a2b/.github/workflows/build.yml#L101

I will test the windows build tomorrow. 

Hey, if it builds and works, you are done!

dail8859:
@diegoiast Appreciate your input!

> I don't see lupdate being run. Does CMake do it for you?

There is a workflow that runs lupdate on a schedule. This is how its always been to keep the .ts files up to date.

https://github.com/dail8859/NotepadNext/blob/e9149614fee2a782c2c751c0159c7cee8bcdf1b3/.github/workflows/lupdate.yml

> I think you nuked translations

I committed this on the wrong branch but it will be needed. Once this moves to Crowdin (see https://github.com/dail8859/NotepadNext/issues/937) then only the English ts file will need updated by lupdate and Crowdin will update the other ts files.

During compile time, cmake does convert the ts files into qm files that are embedded in the executable.

> I do see an increase of 4MB between v0.12 and v0.13. You might want to look at it.

Good to know. I'll need to spin up a VM and figure out how to build my code on Linux and look int it more 🙃

> I see you opted to bring all the code into your repo, instead of using FetchContent as I was aiming to. Why was that?

I am using `CPMAddPackage()` (see https://github.com/dail8859/NotepadNext/blob/e9149614fee2a782c2c751c0159c7cee8bcdf1b3/thirdparty/CMakeLists.txt) on a few repositories currently (waaaay better than submodules) and likely will try to use it more in the future with more repositories.

> What made be not finish the port back then, was the way qmake handled the version.rc or similar parts of the installer. I was hoping no to go the route you went - by using a post-processed file :)

I was hoping to not need this either, but it seems it is part of the build process that qmake just handled completely behind the scenes. It seems very common for applications to do so at least it is in line with many other projects.

> You might want to use a different naming scheme for non official builds (using the git sha1 instead of 0.13)

That's probably a good idea. That way there is only _one_ v0.13 and the rest would have the version show up as the hash.

zocker-160:
> This PR will also mean updating the FlatHub build commands here: https://github.com/flathub/com.github.dail8859.NotepadNext/blob/cfc8b07a9974b78b8163d7cb8d5452d61b06ad23/com.github.dail8859.NotepadNext.yml

@dail8859 you have to keep in mind that during build step there is no network available. The build script downloading anything during compilation will fail.

dail8859:
@zocker-160 That's unfortunate but I've found git submodules to be a huge pain, and manually copying other sources into the repo to be tedious and cumbersome.

I'm open to suggestions but at this point likely won't get rid of the CPMAddPackage() calls due to the ease of managing dependencies.

zocker-160:
@dail8859 understandable, I am just saying and I am not sure if there is a workaround though.

diegoiast:
I added a cmake configuration switch which makes the build system cobsune all 3rd parties from lib instead of cpm.

Then I added a script that parses the main cmake for those projects and pulls them. No ideal. Still working on it.

I would like to port fully to connan to overcome this.

https://github.com/codepointerapp/codepointer/blob/main/get-code.sh

PR #984, opened 2026-03-20
Mark across word boundaries (case sensative) and limit the selection size

### LINKED ISSUE #978: Highlight/Mark entire selection and do not end on word boundaries or non-alpha characters
### Description

Using v0.13 on MacOS with example text:
aa,bb,cc
aa,b,cc

I use my cursor to highlight 'aa,b' as selected text and mark all using style 1.  Result is only 'aa' in line 1 and 'aa' in line 2 is highlighted/marked.  I want to be able to highlight and mark exactly what is selected verbatim.  

Change the selection and highlight logic to not end on word boundaries or non-alpha characters.  Instead use the entire selection as the match criteria.  In the example above, 'aa,b' should be highlighted in line 1 and line 2.

_No response_

cybermartini:
## Summary
- Mark across word boundaries (case sensative) and limit the selection size

Fixes #978
Continuation of #980

### DIFF
--- a/src/decorators/MarkerAppDecorator.cpp
+++ b/src/decorators/MarkerAppDecorator.cpp
@@ -86,7 +86,7 @@ void MarkerAppDecorator::mark(ScintillaNext *editor, int i)
 
     const QByteArray selText = editor->get_text_range(textStart, textEnd);
     Sci_TextToFind ttf {{0, (Sci_PositionCR)editor->length()}, selText.constData(), {-1, -1}};
-    const int flags = isWholeWord ? (SCFIND_MATCHCASE | SCFIND_WHOLEWORD) : SCFIND_MATCHCASE;
+    const int flags = SCFIND_MATCHCASE;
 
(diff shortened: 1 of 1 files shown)


src/decorators/MarkerAppDecorator.cpp
```diff
@@ -55,25 +55,38 @@ QColor MarkerAppDecorator::markerColor(int i) const
...
+    const int mainSelection = editor->mainSelection();
+    const int selectionStart = editor->selectionNStart(mainSelection);
+    const int selectionEnd = editor->selectionNEnd(mainSelection);
 
-    const int curPos = editor->currentPos();
-    const int wordStart = editor->wordStartPosition(curPos, true);
-    const int wordEnd = editor->wordEndPosition(wordStart, true);
+    // If there is a selection, use it verbatim; otherwise fall back to the word at the cursor
+    int textStart, textEnd;
+    bool isWholeWord = false;
```
dail8859, line 64: I see `isWholeWord` is set but not used. Is it intended to be used as part of the flags? Or just not needed at all any more?

PR #986, opened 2026-03-21
Update packaging script for Linux

dail8859:


dail8859:
@StratsScorpio

With then new changes to the install path, I had to also fix some things with building the AppImage.

Do these chances still work for you?

StratsScorpio:
You've only made the adjustments for the AppImage installation. As far as I can tell from the variables (`APPDIR`* instead of `INSTALL_DIR`), this is simply the correct default for compiling to AppImage with CMake.

The native installation part isn't affected. Personally, I only compile natively as long as I can avoid using AppImage (old school :wink: ).

But I can't see anything wrong with the AppImage build itself:

You have `AppDir/usr/bin` and `AppDir/usr/share`, and `AppDir` as the root directory. Looks good :+1:.

PR #987, opened 2026-03-22
PCRE2 Regex Engine

dail8859:
Now with 100% more PCRE2 support


dail8859:
This is a start, but additional testing is needed.

PR #1003, opened 2026-04-16
removed --no-quarantine message

aoplus:
This error is coming while installing on MacOS 26.1 using Brew Error: Calling the `--[no-]quarantine` switch is disabled! There is no replacement.


dail8859:
Thanks! I'm not a Mac person so I'm fine with this.

PR #1007, opened 2026-04-18
feat: Add quick tabs controls to toolbar

ok-developer:
## Description

Added quick actions for tabs manipulation: add new, list existed and close current. Same as in Notepad++.

Screenshot_20260418_224317
Screenshot_20260418_224458
Screenshot_20260418_225859

## Technical Details

- Uses all existed actions: `MainWindow::newFile()`, `MainWindow::closeCurrentFile()` and tabs listing nearly same workflow from `CDockAreaTitleBar::onTabsMenuAboutToShow()`;
- Buttons visibility options for future customization(if else);
- 3 extra icons added from Lucide community;
- Tabs listing button disabled in `DockManager` via config flag.


dail8859:
Hi @ok-developer! Thanks for this PR.

Personally, I've never been a fan of having these options in Notepad++ since there's other ways to do most of this behavior, however I know many other people find it useful, and this code really just reuses mostly existing logic and keeps everything self-contained. Plus it can be configurable later.

So my first few initial asks are:
1. Remove the use of `TabsQuickActionsBar::Private` as nothing else in the code base uses a private struct since there's no intention to expose these classes externally and needing to keep a stable ABI.
2. I'd just ask to try to keep the formatting similar to the existing code. This mostly a failure on my part as I have no clear guidance or tools to enforce a particular style, but for now just use the existing code and keep it fairly similar.
3. Also, I think for any active editors I'd request making the QAction text bold rather than just pretending it with "> " so that it is more visual distinct.

But overall I think the implementation makes sense keeping nearly everything to that single `TabsQuickActionsBar` and the icons are clean and simple.

The other thing I always wrestle with is where to put the 'glue' logic. Right now, the `MainWindow` knows about `TabsQuickActionsBar` implementation but in theory the main window could be passed into the `TabsQuickActionsBar` constructor and let that do all the connecting...I'm fine with the current implementation though...just a thought.

ok-developer:
@dail8859, You're right: it's a small improvement, but the layout of these buttons is already stuck in muscle memory. The first time, I was even a bit confused when I didn't find them)

Updated.

image

dail8859:
Overall looks good. I cleaned it up a bit but kept all the functionality.

Thanks for the PR!

PR #1012, opened 2026-04-24
feat: Preferences dialog view upgrade

ok-developer:
## Description

The preferences dialog has been updated: categories have been added, and the interface has been refreshed. It doesn't really match the spirit of the original Notepad++, but that's actually a plus.

## Overview

<details>
  <summary>Before</summary>

Old view

</details>

<details>
  <summary>After</summary>

https://github.com/user-attachments/assets/df7b751f-723a-482a-be75-3f125614a59e

image
image
  
</details>

## Technical Details

- new settings category addition requires to create subclass of `PreferencesCategoryItem`(which is just item for list model) and add to category list. It provides title, icon path and category settings view realization;
- settings changing works in 'preview' mode: while everything changes on the fly, they can be restored or reset to default values;
- copy and compare functionary have been added to `ApplicationSettings` using Qt meta-object engine(so it's 'automated' and no specific actions needed for new settings fields);
- ~~Russian and Ukrainian translations related to Preferences have also been updated.~~


dail8859:
> new settings category addition requires to create subclass of PreferencesCategoryItem(which is just item for list model) and add to category list. It provides title, icon path and category settings view realization;

Why exactly are custom `PreferencesCategoryItem`s needed? It appears all the UI code was removed from the QCreator's xml format to code in each of the 'categories'...which makes it much harder to maintain/design.

I very quickly mocked up a working example similar to this design (not perfect...I spent like a total of 10 minutes) using a QListWidget and QStackedWidget and literally 0 code changes. Is there some other functionality the abstraction and subclassing provides?

image

PR #1020, opened 2026-04-28
Update portuguese translations

hugok79:
The `pt` lang was added since the last release... there should only be `pt_BR` and `pt_PT` lang codes Nevertheless, `pt` stands for Portuguese language, not Portuguese from Brazil


hugok79:
I just noticed that the translations are through Crowdin...

PR #1038, opened 2026-05-02
Fix file rename: pre-fill dialog and actually delete old file

### LINKED ISSUE #1034: Rename bug
### Operating System and Version

macOS 26.4

### Distribution

brew

### Description

Click on the tab title and then "Rename" doesnt bring the current name in the "Save as" input, and when you rename it the same file remains opened (at least it appears to the user) but in the file system a new one with the new name appears.

### Steps to Reproduce

1. Right Click on tab 
2. click rename
3. Save

### Debug Information

Notepad Next v0.13 
Build Date/Time: Jan 25 2026 20:14:11
Qt: 6.5.3
OS: macOS 26.4
Locale: en_BR
CPU: arm64
File Path: /Applications/Notepadnext.app/Contents/MacOS/NotepadNext
Arguments: /Applications/Notepadnext.app/Contents/MacOS/NotepadNext
Config File: /Users/felipecc/.config/NotepadNext/NotepadNext.ini

### Additional Details

e.g. screenshots, logs, etc

saberoueslati:
Fixes #1034

### Problem

Two bugs caused the rename feature (right-click tab → Rename) to misbehave:

1. **Dialog opened with no pre-filled name.** `MainWindow::renameFile()` passed only the default directory to `FileDialogHelpers::getSaveFileName`. `QFileDialog` only pre-fills the filename field when given a full file path, so the field was always empty.

2. **Old file was never deleted.** `ScintillaNext::rename()` checked `if (saveCopyAs(newFilePath))` to decide whether to remove the original. `saveCopyAs` returns `QFileDevice::FileError`, where `NoError == 0` (falsy). The condition was inverted — it entered the block only on *error*, so on success the `QFile::remove` was skipped and the old file was left on disk.

The combination made it appear as if the file was "renamed" in the tab but the original remained untouched and a second copy appeared at the destination.

Video of the manual test below :

[Screencast from 2026-05-02 00-56-33.webm](https://github.com/user-attachments/assets/c8ae800a-e758-4640-b0a6-1967ac13627c)

### Changes

Pass `editor->getFilePath()` (the current file's full path) to `getSaveFileName` instead of the default directory. `QFileDialog` now navigates to the file's directory and pre-fills its name.

**`src/ScintillaNext.cpp`**

Change the save-success check from `if (saveCopyAs(newFilePath))` to `if (saveCopyAs(newFilePath) == QFileDevice::NoError)`. The old file is now correctly removed after a successful copy.

### DIFF
--- a/src/ScintillaNext.cpp
+++ b/src/ScintillaNext.cpp
@@ -456,7 +456,7 @@ bool ScintillaNext::rename(const QString &newFilePath)
     emit aboutToSave();
 
     // Write out the buffer to the new path
-    if (saveCopyAs(newFilePath)) {
+    if (saveCopyAs(newFilePath) == QFileDevice::NoError) {
         // Remove the old file
         const QString oldPath = fileInfo.canonicalFilePath();
         QFile::remove(oldPath);
--- a/src/dialogs/MainWindow.cpp
+++ b/src/dialogs/MainWindow.cpp
@@ -1479,7 +1479,7 @@ void MainWindow::renameFile()
     if (editor->isFile()) {
         const QString filter = app->getFileDialogFilter();
         QString selectedFilter = app->getFileDialogFilterForLanguage(editor->languageName);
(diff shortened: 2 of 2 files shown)


dail8859:
Awesome! Thank you!

PR #1049, opened 2026-05-11
Add Search and bookmark lines

### LINKED ISSUE #676: [Feature Request] More enhancements with what we can do with bookmarks (including Find to Bookmark)
### Description

In Notepad++ there is a Mark tab, which I have found very useful for certain situations. It is an underrated feature of Notepad++ which does not exist on any other text editor on planet (as far as I know).

image

The checkbox "Bookmark line" lets you also bookmark the lines that are being marked, while "Purge for each search" will remove existing markings and bookmarks.

The ability to bookmark the marked lines comes in very handy when you consider what we can do with bookmarks:

Screenshot 2024-12-06 at 1 58 58 PM

The relevant ones being:
* Cut Bookmarked Lines
* Copy Bookmarked Lines
* Remove Bookmarked Lines
* Remove Unmarked Lines

I would like to see a full-fledged Mark tab in NotepadNext which replicates the function in Notepad++.

If a full-fledged "Mark" tab is not planned right now, perhaps the "Find" tab in NotepadNext could have a button to bookmark by searching:

Screenshot 2024-12-06 at 2 05 01 PM

Then the Bookmarks menu can be enhanced with more options:

Screenshot 2024-12-06 at 2 09 19 PM

Not sure if there's an alternative for this is NotepadNext right now. If yes, willing to discuss.

### LINKED ISSUE #829: Search and bookmark lines
### Description

Hi,
First of all, thank you for this great project, NotepadNext is a very promising alternative to Notepad++ on macOS.

One of the most useful features I use regularly in Notepad++ is the ability to automatically bookmark all lines that match a search pattern via the Mark tab. This allows quick navigation, selection, copy, cut or removal of all matching lines.

In Notepad++, this feature is enabled by checking the box "Bookmark line" before clicking "Mark All". It is especially useful for batch processing or filtering content.

Currently, NotepadNext doesn’t seem to support this functionality. I would love to see it implemented in a future release.

Thanks again for your great work!

_No response_

_No response_

saberoueslati:
## Description

Fixes #829

This PR adds a **Mark** tab to the Find/Replace dialog so users can highlight all lines matching a search pattern directly in the editor — mirroring the Mark tab feature in Notepad++.

The tab uses a dedicated Scintilla indicator to highlight matched text in-place. An optional **Bookmark line** checkbox also places margin bookmarks on each matching line. A **Purge for each search** checkbox clears previous highlights before applying new ones, while **Clear all marks** and **Copy Marked Text** complete the workflow. A new **Search and Bookmark Lines…** menu entry under Search → Bookmarks opens the dialog directly at the Mark tab.

---

## Changes Made

**`src/decorators/BookMarkDecorator.h` / `.cpp`**
- `addBookmark(int line)` now returns the Scintilla marker handle (`int`) instead of `void`, enabling handle-based removal.
- Added `removeBookmarkHandle(int markerHandle)` — removes a bookmark by its handle rather than by line number.

**`src/dialogs/FindReplaceDialog.h`**
- Dropped the unused `FIND_IN_FILES_TAB = 2` enum value; `MARK_TAB` is now `2`.
- Replaced `bookmarkMatchingLines()` with three new public slots: `markAll()`, `clearAllMarks()`, `copyMarkedText()`.
- Added private `int markIndicator = -1;` and `void ensureMarkIndicator();` to manage the per-editor indicator.

**`src/dialogs/FindReplaceDialog.ui`**
- Added `buttonMarkAll` ("Mark All"), `buttonClearAllMarks` ("Clear all marks"), and `buttonCopyMarkedText` ("Copy Marked Text") to the button column.
- Added `checkBoxBookmarkLine` and `checkBoxPurgeForEachSearch` to the options column; both are hidden in the Find and Replace tabs and shown only in the Mark tab.
- Updated `<tabstops>` accordingly.

**`src/dialogs/FindReplaceDialog.cpp`**
- Adds a "Mark" third tab to `tabBar` in the constructor.
- `changeTab()`: full three-way tab management — the Mark tab hides Find/Replace/Count/FindAll buttons and shows the three Mark buttons plus the two Mark-only checkboxes; switching back to Find or Replace restores their buttons and hides the Mark-only widgets. The Find button is hidden while the Mark tab is active.
- `ensureMarkIndicator()`: allocates a `find_mark_highlight` indicator on the current editor (golden, `INDIC_FULLBOX`) and applies style. Safe to call on every `markAll()` — same name returns the same handle per editor.
- `setEditor()`: resets `markIndicator = -1` on editor switch so the indicator is re-styled on the new editor.
- `markAll()`: calls `prepareToPerformSearch()`, fills the indicator for every match via `finder->forEachMatch()`, optionally purges first, and optionally places line bookmarks; shows "Mark: N matches in entire file".
- `clearAllMarks()`: calls `indicatorClearRange` over the full document.
- `copyMarkedText()`: walks indicator runs with `indicatorValueAt` / `indicatorEnd`, collects matched text, and writes newline-joined results to the clipboard.

**`src/dialogs/MainWindow.ui`**
- Added `actionSearchAndBookmark` to `menuBookmark` immediately after `actionToggleBookmark`.
- Added the action definition with text "Search and Bookmark Lines…".

- Connected `actionSearchAndBookmark` to `showFindReplaceDialog(FindReplaceDialog::MARK_TAB)`.

---

## Testing

Manual test plan: 

[testing-issue-829-search-bookmark.md](https://github.com/user-attachments/files/27575899/testing-issue-829-search-bookmark.md)

Small manual test of the feature in the video below :

[Screencast from 2026-05-11 01-06-38.webm](https://github.com/user-attachments/assets/e07a7981-9c57-4422-bc33-f8eb43d2d093)

Coverage includes:

- Mark tab is visible and clickable; correct buttons shown/hidden per tab.
- Mark All highlights matched text with a golden indicator in the editor.
- Clear all marks removes all highlights.
- Bookmark line checkbox adds margin bookmarks alongside highlights.
- Purge for each search clears previous marks before applying new ones.
- Without Purge, successive Mark All calls accumulate marks from different search terms.
- Copy Marked Text copies all highlighted ranges to the clipboard, newline-separated.
- Copy Marked Text with no marks shows an informative message without crashing.
- Zero-match search shows "Mark: 0 matches in entire file".
- Menu entry "Search and Bookmark Lines…" opens the dialog at the Mark tab.
- Tab switching shows and hides the correct controls on every transition.
- Regex search highlights only the matching ranges.
- Mark All does not disturb manually placed bookmarks.
- Marks survive editor tab switching; Clear all marks and Copy Marked Text work after switching back.
- Regex mode keeps Backward direction disabled correctly across tab switches.

### DIFF
--- a/src/decorators/BookMarkDecorator.cpp
+++ b/src/decorators/BookMarkDecorator.cpp
@@ -42,14 +42,9 @@ bool BookMarkDecorator::isBookmarkSet(int line) const
     return editor->markerGet(line) & (1 << MARK_BOOKMARK);
 }
 
-int BookMarkDecorator::addBookmark(int line)
+void BookMarkDecorator::addBookmark(int line)
 {
-    return editor->markerAdd(line, MARK_BOOKMARK);
-}
-
-void BookMarkDecorator::removeBookmarkHandle(int markerHandle)
-{
-    editor->markerDeleteHandle(markerHandle);
+    editor->markerAdd(line, MARK_BOOKMARK);
 }
 
 void BookMarkDecorator::removeBookmark(int line)
--- a/src/decorators/BookMarkDecorator.cpp
+++ b/src/decorators/BookMarkDecorator.cpp
@@ -42,9 +42,14 @@ bool BookMarkDecorator::isBookmarkSet(int line) const
     return editor->markerGet(line) & (1 << MARK_BOOKMARK);
 }
 
-void BookMarkDecorator::addBookmark(int line)
+int BookMarkDecorator::addBookmark(int line)
 {
-    editor->markerAdd(line, MARK_BOOKMARK);
+    return editor->markerAdd(line, MARK_BOOKMARK);
+}
+
+void BookMarkDecorator::removeBookmarkHandle(int markerHandle)
+{
+    editor->markerDeleteHandle(markerHandle);
 }
 
 void BookMarkDecorator::removeBookmark(int line)
--- a/src/decorators/BookMarkDecorator.h
+++ b/src/decorators/BookMarkDecorator.h
@@ -30,8 +30,7 @@ public:
     explicit BookMarkDecorator(ScintillaNext *editor);
 
     bool isBookmarkSet(int line) const;
-    int addBookmark(int line);
-    void removeBookmarkHandle(int markerHandle);
+    void addBookmark(int line);
     void removeBookmark(int line);
     void toggleBookmark(int line);
 
--- a/src/decorators/BookMarkDecorator.h
+++ b/src/decorators/BookMarkDecorator.h
@@ -30,7 +30,8 @@ public:
     explicit BookMarkDecorator(ScintillaNext *editor);
 
     bool isBookmarkSet(int line) const;
-    void addBookmark(int line);
+    int addBookmark(int line);
+    void removeBookmarkHandle(int markerHandle);
     void removeBookmark(int line);
     void toggleBookmark(int line);
 
--- a/src/dialogs/FindReplaceDialog.cpp
+++ b/src/dialogs/FindReplaceDialog.cpp
@@ -24,9 +24,12 @@
 #include <QStatusBar>
 #include <QLineEdit>
 #include <QKeyEvent>
+#include <QClipboard>
+#include <QGuiApplication>
 
 #include "ScintillaNext.h"
 #include "MainWindow.h"
+#include "BookMarkDecorator.h"
 
 
 static void convertToExtended(QString &str)
@@ -58,6 +61,7 @@ FindReplaceDialog::FindReplaceDialog(ISearchResultsHandler *searchResults, MainW
     tabBar = new QTabBar();
     tabBar->addTab(tr("Find"));
     tabBar->addTab(tr("Replace"));
+    tabBar->addTab(tr("Mark"));
     tabBar->setExpanding(false);
     qobject_cast<QVBoxLayout *>(layout())->insertWidget(0, tabBar);
     connect(tabBar, &QTabBar::currentChanged, this, &FindReplaceDialog::changeTab);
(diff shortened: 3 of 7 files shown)


dail8859:
Hi @saberoueslati

Overall I think this looks pretty good. My one question is why the need for storing the handles in `markedBookmarkHandles` to delete each one individually? Offhand I would think that `decorator->clearAllBookmarks();` would be sufficient, but maybe there's a reason I don't see initially.

saberoueslati:
@dail8859 that approach was chosen because `decorator->clearAllBookmarks()` is broad: it deletes every bookmark in the editor, including bookmarks the user placed manually before running Mark All.

markedBookmarkHandles lets us delete only the bookmark markers created by this Mark All feature and leaves the bookmarks placed by the user manually alone, that's the idea here.

If this behavior is overkill or unnecessary to you, I can just replace it with the `clearAllBookmarks` decorator.

dail8859:
Ah ok. I checked Notepad++ next behavior, and it simply deletes all bookmarks of that type -- no matter how they were created. I think this is the best approach to take for now as it reduces complexity and removes the extra data structure to track handles.

saberoueslati:
@dail8859 sure, I'll make a commit for the change shortly.

saberoueslati:
@dail8859 Done, now `decorator->clearAllBookmarks()` is used

dail8859:
I pushed a couple minor changes. I removed storing the `markIndicator` indicator and it can always be retrieved with `ensureMarkIndicator()`. Yes in theory it does not need to configure the indicator style each time it is called but those should be incredibly low overhead and simplify things. There are also no tight loops that call `ensureMarkIndicator()` so I'm not worried.

Also I added the `Ctrl+M` shortcut since that's what Notepad++ uses. I thought about moving it to a similar location as Notepad++ and naming it "Mark..." but I think it makes sense and clearer where it is in this PR.

dail8859:
Thanks again!

PR #1053, opened 2026-05-17
Gemini markdown formatting

Rambo19911:


dail8859:
There are too many changes to review.

Also with that there's no details what this is intending to do exactly. The gemini API is not desirable to ahve for this application since only a very small fraction of users would use this functionality.

PR #1062, opened 2026-05-23
Add session snapshot feature to preserve unsaved files on exit

### LINKED ISSUE #81: [Feature Request] Close without saving modified files
The default behaviour in Notepad++ has always been to retain any modified buffers when closed without prompting the user to save. This would be a useful feature to have in order to use NotepadNext as a scratchpad of sorts.

Thanks,
digg33

### LINKED ISSUE #178: feature: Auto save & load unnamed documents
Just like notepad++ doing, so we can use it as a very handy scribbling pad.

thanks.

chen3feng:
## Summary

This PR implements Notepad++'s session snapshot feature — unsaved "New N" files and modified files are automatically preserved on exit and restored on next launch. No more data loss from accidental closes.

## Changes

- Add `EnableSessionSnapshot` setting (default **on**) as a master safety-net toggle
- Change all three session restore settings defaults to `true`
- Always save `UnsavedFile` and `TempFile` content regardless of restore settings (safety net)
- Filter session **loading** based on per-type restore preferences
- Reduce auto-save interval from 60s to **7s** (matching Notepad++)
- Trigger session backup on editor tab switches for extra safety
- Remove "experimental" warning from Preferences dialog
- Add "Previously open files" restore checkbox to Preferences UI

## Behavior

- **Default**: close without save prompts, content auto-preserved, restored on relaunch
- **Opt-out**: uncheck "Enable Session Snapshot" in Preferences → reverts to old behavior with save prompts
- **Fine control**: keep master ON, uncheck individual restore options (e.g., don't restore temp files but still back them up)

## Files Changed

| File | Change |
|------|--------|
| `src/ApplicationSettings.h` | Add `EnableSessionSnapshot` setting |
| `src/ApplicationSettings.cpp` | New default `true`; all restore defaults → `true` |
| `src/NotepadNextApplication.cpp` | Snapshot gate; timer 60→7s; tab-switch backup; initialize session manager before load |
| `src/SessionManager.cpp` | Always save UnsavedFile/TempFile; filter load by flags; `willFileGetStoredInSession` unconditional for unsaved/temp |
| `src/dialogs/PreferencesDialog.cpp` | Wire master switch; remove experimental warning; enable/disable logic |
| `src/dialogs/PreferencesDialog.ui` | Add `checkBoxSessionSnapshot`; restructure restore group box |

Closes #81, Closes #178

### DIFF
--- a/src/ApplicationSettings.cpp
+++ b/src/ApplicationSettings.cpp
@@ -51,9 +51,10 @@ CREATE_SETTING(Gui, ExitOnLastTabClosed, exitOnLastTabClosed, bool, false)
 
 CREATE_SETTING(Gui, CombineSearchResults, combineSearchResults, bool, false)
 
-CREATE_SETTING(App, RestorePreviousSession, restorePreviousSession, bool, false)
-CREATE_SETTING(App, RestoreUnsavedFiles, restoreUnsavedFiles, bool, false)
-CREATE_SETTING(App, RestoreTempFiles, restoreTempFiles, bool, false)
+CREATE_SETTING(App, EnableSessionSnapshot, enableSessionSnapshot, bool, true)
+CREATE_SETTING(App, RestorePreviousSession, restorePreviousSession, bool, true)
+CREATE_SETTING(App, RestoreUnsavedFiles, restoreUnsavedFiles, bool, true)
+CREATE_SETTING(App, RestoreTempFiles, restoreTempFiles, bool, true)
 
 CREATE_SETTING(App, DefaultDirectoryBehavior, defaultDirectoryBehavior, ApplicationSettings::DefaultDirectoryBehaviorEnum, ApplicationSettings::FollowCurrentDocument)
 CREATE_SETTING(App, DefaultDirectory, defaultDirectory, QString, QString())
--- a/src/ApplicationSettings.h
+++ b/src/ApplicationSettings.h
@@ -99,6 +99,7 @@ public:
 
     DEFINE_SETTING(CombineSearchResults, combineSearchResults, bool)
 
+    DEFINE_SETTING(EnableSessionSnapshot, enableSessionSnapshot, bool)
     DEFINE_SETTING(RestorePreviousSession, restorePreviousSession, bool)
     DEFINE_SETTING(RestoreUnsavedFiles, restoreUnsavedFiles, bool)
     DEFINE_SETTING(RestoreTempFiles, restoreTempFiles, bool)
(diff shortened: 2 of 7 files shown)


chen3feng:
The `build (windows-latest, 6.5)` failure is **unrelated to this PR**. All other CI jobs pass, including Windows with Qt 6.8 and 6.10.

**Root cause**: GitHub recently upgraded `windows-latest` from `windows-2022` to `windows-2025-vs2026`. The new MSVC 2026 compiler removed `stdext` types that Qt 6.5's `qvarlengtharray.h` depends on:

```
error C3861: 'stdext': identifier not found
  → qvarlengtharray.h(379)
```

This is a pre-existing incompatibility between Qt 6.5 and the new compiler toolchain, not caused by any changes in this PR.

**Fix options** (for a separate PR):
1. Pin the Windows 6.5 job to `windows-2022` explicitly in the CI matrix
2. Drop Qt 6.5 Windows CI since Qt 6.8+ builds fine

dail8859:
> This PR implements Notepad++'s session snapshot feature — unsaved "New N" files and modified files are automatically preserved on exit and restored on next launch. 

Maybe I'm a bit confused here...but doesn't the application have this capability already? The two issues you referenced are closed. You can restore the previous session along with unsaved files and temp files.

What exact set of steps demonstrate the functionality added.

chen3feng:
You're right — the underlying SessionManager capability already exists. This PR doesn't add a new mechanism; it makes the existing one **work out of the box** and removes the friction that keeps most users from ever discovering it.

On current `master`:
- All three restore settings (`RestorePreviousSession`, `RestoreUnsavedFiles`, `RestoreTempFiles`) default to **off**
- The feature is buried in Preferences behind an **"experimental … may lead to possible data loss"** warning dialog
- So out of the box, closing the app with an unsaved "New 1" tab still loses it

Steps that show the difference:
1. Fresh install (default settings), open the app
2. Type into the default "New 1" tab without saving
3. Quit (Cmd+Q / close window)
4. **master**: prompted to save, lose the content if you don't → reopen shows an empty "New 1"
   **this PR**: closes silently, reopen restores "New 1" with its content — like Notepad++

What the PR actually changes:
- Flip the three restore defaults to **on**
- Remove the "experimental" warning
- Reduce the autosave interval 60s → 7s (matching Notepad++)
- Back up the active editor on tab switches
- Add a master "Enable Session Snapshot" toggle so anyone who dislikes this behavior can turn the whole thing off in one click (and the per-type restore checkboxes remain for finer control)

So it's really a UX/defaults change rather than new functionality. Totally understand if you'd prefer to keep it opt-in by default — happy to drop the default-on part and keep just the autosave-interval / tab-switch-backup / de-"experimental" improvements if that's more in line with your intent.

PR #1068, opened 2026-05-30
Add auto-completion toggle setting

### LINKED ISSUE #1025: Auto Completion Toggle?
### Description

The Auto Completion is set to be automatically on, and I can't find where to turn it off. Will a toggle of turning on / off "Auto Completion" been added? Or have I missed something?

_No response_

_No response_

saberoueslati:
## Description

Fixes #1025

Auto-completion is enabled by default in NotepadNext with no way to turn it off. This PR adds a persistent **Auto Completion** toggle so users can disable it, exposed both in the Preferences dialog and the View menu. The setting defaults to `true`, preserving existing behaviour for all current users.

---

## Changes

**`src/ApplicationSettings.h` / `.cpp`**
- Added `AutoCompletion` bool setting (stored as `Editor/AutoCompletion`, default `true`) using the existing `DEFINE_SETTING` / `CREATE_SETTING` macro system.

**`src/EditorManager.cpp`**
- Connected `autoCompletionChanged` signal in the constructor — iterates all open editors, enables/disables the `AutoCompletion` decorator, and calls `autoCCancel()` to dismiss any visible popup when toggling off.
- Changed `ac->setEnabled(true)` in `setupEditor` to `ac->setEnabled(settings->autoCompletion())` so newly opened files respect the stored preference.

**`src/dialogs/PreferencesDialog.ui` / `.cpp`**
- Added an **Auto Completion** checkbox to the editor settings section, mapped to the new setting via `MapSettingToCheckBox`.

**`src/dialogs/MainWindow.ui`**
- Defined `actionAutoCompletion` as a checkable action with the text "Auto Completion".
- Added it to the View menu immediately after **Word Wrap**.

- Wired `actionAutoCompletion` with a three-way bidirectional sync: initial state set from the stored setting, View menu reflects changes made in Preferences, and Preferences reflects changes made from the View menu.

Video of manual test below :

[Screencast from 2026-05-30 14-21-37.webm](https://github.com/user-attachments/assets/fcf38bd7-bb4b-4cfc-843b-f4dbce79ea59)

### DIFF
--- a/src/ApplicationSettings.cpp
+++ b/src/ApplicationSettings.cpp
@@ -71,3 +71,4 @@ CREATE_SETTING(Editor, AdditionalWordChars, additionalWordChars, QString, QStrin
 CREATE_SETTING(Editor, DefaultEOLMode, defaultEOLMode, QString, QStringLiteral(""))
 CREATE_SETTING(Editor, URLHighlighting, urlHighlighting, bool, true)
 CREATE_SETTING(Editor, ShowLineNumbers, showLineNumbers, bool, true)
+CREATE_SETTING(Editor, AutoCompletion, autoCompletion, bool, true)
--- a/src/ApplicationSettings.h
+++ b/src/ApplicationSettings.h
@@ -119,4 +119,5 @@ public:
     DEFINE_SETTING(DefaultEOLMode, defaultEOLMode, QString)
     DEFINE_SETTING(URLHighlighting, urlHighlighting, bool)
     DEFINE_SETTING(ShowLineNumbers, showLineNumbers, bool)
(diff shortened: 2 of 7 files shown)


dail8859:
Thanks @saberoueslati I think this looks great!

The one thing I changed was I went ahead and removed it in the MainWindow menu...I wouldn't foresee that a user would need to toggle it on and off very frequently. It can always be added back in later if needed of course. But for now I think a preference option is good enough.

bradsmr:
Thanks for adding this feature! Any timeline on when this will make it into the next MacOS release?

PR #1069, opened 2026-06-02
[DEMO] Linux native packages with repositories on GH actions and S3

olegantonyan:
Hello
I made a software that might be useful for NotepadNext - https://omnipackage.org

It wraps native Linux tools to make RPM&DEB packages and manages repositories on S3.

For users the entrypoint could look like this - https://repositories-test.omnipackage.org/notepadnext/install.html (static page generated after build and put in S3 along with repo files). 
Choose distro, copy-paste commands and install the software. From now on they'll pick up updates from you automatically and natively through distro's package manager as if notepadnext was in distro's standard repos.

It runs entirely in Github Actions (but not limited to - it's just a CLI that needs Linux with containers). You bring own S3-compatible bucket and GPG key. The demo linked above was built in my fork (https://github.com/olegantonyan/NotepadNext/actions/runs/26813366189) and hosted in my Cloudlflare R2 bucket (which is nice for indie repos with free egress and 10G storage, but requires custom domain).

This PR is just a demo and not supposed to be merged. If you decide to pull the trigger on it we'd need to make a few decisions and improvements, like stable vs dev builds, containers caching, distro support, etc. Feel free to close this PR, or give any or none feedback in general.


dail8859:
Appreciate the info. I'll keep it in mind. Deploying code on Linux is definitely a huge headache, and it seems any direction you go there's always some weird process to get alot of cross compatibility.

PR #1070, opened 2026-06-02
Add language keyword candidates to auto-completion

### LINKED ISSUE #839: Language keyword auto-completion does not work (Linux)
### Operating System and Version

Fedora 41

### Distribution

flathub

### Description

When typing, keyword suggestions should appear based on the language you are in.

In C++, when typing "publ", the autocompletion should suggest "public" based on the keywords set here: https://github.com/dail8859/NotepadNext/blob/master/src/NotepadNext/languages/cpp.lua

In every language the keyword autocompletion is broken. The autocomplete only suggests existing words in your file.

### Steps to Reproduce

1. Create a new file test.cpp.
2. Open in NotepadNext
3. Type in any keyword. For example, "class" or "public". You will notice no suggestions appear.
4. Once you have words in the file, suggestions will appear, but only those existing in the file already.

### Debug Information

Notepad Next v0.12  (Flatpak)
Build Date/Time: Nov 11 2011 11:11:11
Qt: 6.5.3
OS: KDE Flatpak runtime
Locale: en_US
CPU: x86_64
File Path: /app/bin/NotepadNext
Arguments: NotepadNext
Config File: /home/cory/.var/app/com.github.dail8859.NotepadNext/config/NotepadNext/NotepadNext.ini

### Additional Details

Image

Image

saberoueslati:
## Description

Fixes #839

NotepadNext's auto-completion only suggests words already present in the document. This PR adds language keyword candidates to the completion list so keywords appear as suggestions even before they have been typed in the file. The change is gated by the existing **Auto Completion** preference and requires no new settings.

---

## Changes

**`src/ScintillaNext.h`**
- Added `#include <QStringList>` and a public `QStringList languageKeywords` member to hold the active language's keyword list alongside the existing `languageName` and `languageSingleLineComment` fields.

**`src/scripts/init.lua`**
- Added `GetLanguageKeywords(languageName)` helper that reads `languages[languageName].keywords`, splits each space-separated keyword string into individual tokens, deduplicates them, recurses into `additionalLanguages` (matching `SetLanguage`'s behaviour), and returns a sorted array. Returns `{}` for languages with no keyword definitions (e.g. plain Text).

**`src/NotepadNextApplication.cpp`**
- Added one line at the end of `setEditorLanguage()` — calls `GetLanguageKeywords(languageName)` through the Lua state and stores the result in `editor->languageKeywords`. The field is unconditionally reassigned on every language change, so switching to Text or any other keyword-less language clears it automatically.

**`src/decorators/AutoCompletion.cpp`**
- Extended `showAutoCompletion()` to iterate `editor->languageKeywords` after the document scan, inserting any keyword that starts with the current prefix into the existing `QSet<QByteArray>` candidate set. Deduplication is handled for free by the set. The 3-character minimum, `SCFIND_MATCHCASE` case-sensitivity, `SC_ORDER_PERFORMSORT` ordering, and the single `Editor/AutoCompletion` preference toggle are all preserved.

Video of manual test below :

[Screencast from 2026-06-02 21-34-24.webm](https://github.com/user-attachments/assets/a76cf7ae-181e-4522-8c63-3f2f67693f34)

### DIFF
--- a/src/NotepadNextApplication.cpp
+++ b/src/NotepadNextApplication.cpp
@@ -295,6 +295,7 @@ void NotepadNextApplication::setEditorLanguage(ScintillaNext *editor, const QStr
     getLuaState()->setVariable("skip_tabwidth", skipTabWidth);
 
     getLuaState()->execute("SetLanguage(languageName)");
+    editor->languageKeywords = getLuaState()->executeAndReturn<QStringList>("return GetLanguageKeywords(languageName)");
 }
 
 QString NotepadNextApplication::detectLanguage(ScintillaNext *editor) const
--- a/src/NotepadNextApplication.cpp
+++ b/src/NotepadNextApplication.cpp
@@ -295,7 +295,12 @@ void NotepadNextApplication::setEditorLanguage(ScintillaNext *editor, const QStr
     getLuaState()->setVariable("skip_tabwidth", skipTabWidth);
 
     getLuaState()->execute("SetLanguage(languageName)");
-    editor->languageKeywords = getLuaState()->executeAndReturn<QStringList>("return GetLanguageKeywords(languageName)");
+}
+
+QStringList NotepadNextApplication::getLanguageKeywords(const QString &languageName) const
+{
+    getLuaState()->setVariable("languageName", languageName);
+    return getLuaState()->executeAndReturn<QStringList>("return GetLanguageKeywords(languageName)");
 }
 
 QString NotepadNextApplication::detectLanguage(ScintillaNext *editor) const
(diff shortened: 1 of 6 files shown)


dail8859 (changes requested):
I think the best implementation for now would be to move the `languageKeywords` to be stored in the AutoCompletion class and not in the editor since the AutoCompletion is the only place that uses it for now.

My thought is:
- On AutoCompletion creation (or on the ScintillaNext::lexerChanged signal) call `app->GetLanguageKeywords(lang)` and store the returned value in a private list in AutoCompletion
- `app->GetLanguageKeywords(editor->languageName)` would simply wrap the call to `getLuaState()->executeAndReturn<QStringList>("return GetLanguageKeywords(languageName)");`

That way the keywords don't live in the editor and the application is responsible for pulling lexer information.

saberoueslati:
@dail8859 I'll take care of it, I'll ping you as soon as I'm finished with it.

saberoueslati:
@dail8859 I made the changes as you requested

dail8859:
Apologies for the delay. Thanks for the PR!

PR #1045, opened 2026-05-05
Add Find/Replace In Selection

saberoueslati:
## Description

Fixes #856

This PR adds an **In selection** option to the Find/Replace dialog so users can limit find, count, find-all, replace, and replace-all operations to the selected text range.

The selection range is captured when the checkbox is enabled, so later cursor movement or selection changes do not accidentally expand the search scope.

Additionally, **Mark All** is constrained to the captured selection range when **In selection** is active, and **Purge for each search** clears highlights only within that range rather than the entire document.

## Changes Made

- Added an `In selection` checkbox to the Find/Replace dialog.
- Enable the checkbox only for a valid single stream selection.
- Disable it for no selection, rectangular selections, and multiple selections.
- Capture the selected range when `In selection` is checked.
- Constrain the following operations to the captured selection range:
  - Find Next
  - Find Previous
  - Count
  - Find All in Current Document
  - Replace
  - Replace All
- Keep replacement bounds updated when replacements change text length.
- Prevent single Replace from modifying a match outside the captured selection.
- Ensure Find Previous inside a selection uses the current search flags, including Match Case, Whole Word, and Regex.
- Reset the in-selection state when switching editor tabs.
  - Mark All
- Constrain "Purge for each search" clearing to the captured selection range when In selection is active.
- Show distinct status bar messages: "match in selection" vs "match in entire file" for Mark All.

## Testing

Manual test plan : 
[TestPlan-FindReplaceInSelection.md](https://github.com/user-attachments/files/28318016/TestPlan-FindReplaceInSelection.md)

Coverage includes:

- Checkbox enabled/disabled state for normal, empty, rectangular, and multiple selections.
- Find Next and Find Previous constrained to the captured selection.
- Wrap Around behavior inside the captured selection.
- Count constrained to the captured selection.
- Replace All constrained to the captured selection.
- Replacement length changes updating the selected range correctly.
- Find All in Current Document constrained to the captured selection.
- Editor tab switching clearing the in-selection state.
- Match Case, Whole Word, and Regex behavior with In selection.
- Regression test for Find Previous using stale search flags.
- Regression test for single Replace not modifying matches outside the captured selection.
- Mark All constrained to the captured selection range.
- Purge for each search clearing only within the selection (marks outside survive).
- Status bar message distinguishing in-selection vs entire-file for Mark All.
- Clear All Marks removing all highlights regardless of In selection state.
- Bookmark Line with In Selection bookmarking only lines within the captured range.
[...]

PR #1071, opened 2026-06-07
Fix context menu caret behavior when right-clicking inside a selection

### LINKED ISSUE #780: Editor Context Menu Location

saberoueslati:
Fixes #780

The editor context menu had inconsistent behavior; this meant right-clicking outside a selection with the intent to act on that location would still operate on wherever the caret happened to be.

## Change

**MainWindow.cpp** — In the `customContextMenuRequested` handler inside `addEditor`, a call to `selectionFromPoint` now checks whether the right-click landed inside an existing selection. If it did, the selection and caret are left untouched. If it did not, `setEmptySelection` is called to move the caret to the clicked position, matching standard editor behavior.

## Behavior

- Right-clicking **inside** a selection preserves the selection; context menu actions operate on the selected text
- Right-clicking **outside** a selection clears it and moves the caret to the clicked position
- URL detection (`Copy URL`) is unaffected — `contextMenuPos` is still set before the check
- No new settings, files, or API changes

Manual test after the fix :

[Screencast from 2026-06-07 23-28-54.webm](https://github.com/user-attachments/assets/7f3c752f-f95f-4555-a453-d66f37df6a1a)

### DIFF
--- a/src/dialogs/MainWindow.cpp
+++ b/src/dialogs/MainWindow.cpp
@@ -2105,6 +2105,12 @@ void MainWindow::addEditor(ScintillaNext *editor)
     connect(editor, &ScintillaNext::customContextMenuRequested, this, [=, this](const QPoint &pos) {
         contextMenuPos = editor->positionFromPoint(pos.x(), pos.y());
 
+        // If the click landed outside the current selection, move the caret there
+        // and clear the selection so caret-based actions use the clicked location.
+        if (editor->selectionFromPoint(pos.x(), pos.y()) == -1) {
+            editor->setEmptySelection(contextMenuPos);
+        }
+
         QStringList actionNames = {
             "Cut",
             "Copy",


dail8859:
Thanks! I was not aware Scintilla had the `selectionFromPoint()` API but that's exactly what is needed.

saberoueslati:
Glad to help @dail8859, while you are at it, can you please review this PR https://github.com/dail8859/NotepadNext/pull/1045 ?

Thank you

PR #1072, opened 2026-06-08
Make search result text selectable and copyable

### LINKED ISSUE #825: Ability to Select and Copy Text from search results (Mac)
### Description

At the moment, users are unabel to select text in the search results.  You can copy the whole search data with the copy button. 

Would like to allow users to select certain text in their grep or search results

_No response_

saberoueslati:
## Description

Closes #825

This PR makes text inside Search Results rows selectable and copyable without changing the existing copy-all behavior. Individual result text can now be selected with the mouse and copied as plain text via `Ctrl+C`, `Edit > Copy`, or the right-click context menu. The toolbar copy button still copies all search results.

---

## Changes

**`src/docks/SearchResultsDock.cpp`**
- Added an embedded read-only `QTextEdit` for individual search result rows so result text can be selected directly.
- Added plain-text copy support for selected result text, without copying highlight formatting.
- Added result-row context menu support for `Copy` and `Copy Selected Line`.
- Added `Copy Selected Line` to the tree row context menu so right-clicking the line-number cell also supports copying the full result line.
- Preserved double-click activation so opening/navigating to a search result still works.
- Clears previous result-text selection when left-clicking in Search Results.
- Updates embedded result text widget fonts when the configured editor font changes.
- Kept the existing toolbar copy button behavior for copying all search results.

**`src/docks/SearchResultsDock.h`**
- Added `copySelectedResultText()` so the main window copy action can copy selected Search Results text when appropriate.
- Added event filtering and selection-clearing helpers for result text selection handling.

- Updated `Edit > Copy` to first copy selected Search Results text when a result-text selection exists, then fall back to the current editor copy behavior.

---

## Manual Testing

- Ran a search and selected part of an individual result row.
- Verified `Ctrl+C` copies only the selected result text as plain text.
- Verified `Edit > Copy` copies selected result text even after opening the Edit menu.
- Verified the result text right-click menu `Copy` copies selected text.
- Verified `Copy Selected Line` copies the full result line from the result text context menu.
- Verified `Copy Selected Line` is also available when right-clicking the line-number cell.
- Verified the toolbar copy button still copies all search results.
- Verified double-clicking a result still navigates to the match.

Video of basic manual test below:

[Screencast from 2026-06-08 23-04-01.webm](https://github.com/user-attachments/assets/2df076eb-175b-4074-a3eb-4d511d432676)

### DIFF
--- a/src/dialogs/MainWindow.cpp
+++ b/src/dialogs/MainWindow.cpp
@@ -335,7 +335,16 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
     connectEditorAction(ui->actionUndo, &ScintillaNext::undo);
     connectEditorAction(ui->actionRedo, &ScintillaNext::redo);
     connectEditorAction(ui->actionCut, &ScintillaNext::cutAllowLine);
-    connectEditorAction(ui->actionCopy, &ScintillaNext::copyAllowLine);
+    connect(ui->actionCopy, &QAction::triggered, this, [this]() {
+        SearchResultsDock *searchResultsDock = findChild<SearchResultsDock *>();
+        if (searchResultsDock != Q_NULLPTR && searchResultsDock->copySelectedResultText()) {
+            return;
+        }
+
+        if (auto* editor = currentEditor()) {
+            editor->copyAllowLine();
+        }
+    });
     connectEditorAction(ui->actionDelete, &ScintillaNext::clear);
     connectEditorAction(ui->actionPaste, &ScintillaNext::paste);
     connectEditorAction(ui->actionSelectAll, &ScintillaNext::selectAll);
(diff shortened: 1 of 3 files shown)


dail8859:
I did some testing and noticed a significant lag when the search fines more that ~100 results. Even in release mode it is noticeable. So likely creating a `QTextEdit` for each result is quite tasking. I know Notepad++ has the capabilities of selecting a substring within the result but given the features needed to make this happen (e.g. a QTextEdit per entry) this probably isn't justifiable.

Given that...I think it is feasible that if a user wants _part_ of a line that they can just double click the result and go to that line in the file and copy the text they need. This is very much how other code editors (e.g. Visual Studio) does it.

I tried out some changes real quick and with a bit of code got this to work which I think should suffice in most cases. I'll clean it up and open a PR soon for you to look at:

FindResults

PR #1084, opened 2026-07-03
Fix MacOS build

dail8859:


alsyundawy:
running whtout issue

https://github.com/alsyundawy/NotepadNext-MacOS/actions/runs/28825487607

PR #1086, opened 2026-07-07
Allow multiple search results to be copied

dail8859:
CC: @saberoueslati

This obviously doesn't make it selectable. I would still be open to that if there's a light-weight and non-invasive way to do it. Notepad++ collects all its search results in another text editor that it highly controls so it gives a bit more 'text-like' capabilities.

The other thing was in #1072 was the MainWindow.cpp changes. This is a problem...but I think the issue is larger than a specific check. In theory any `QDockWidget` should be able to add their own shortcuts (e.g. Lua Console, Debug Log, etc) and not require the main window to know about what shortcuts it needs.


saberoueslati:
Sorry for the delay, works as intended when I tested it locally, looks good to me, I'll try reworking a non invasive way for selecting text in individual search results as suggested.

dail8859:
No worries. I've also been super busy lately so I apologize for the delay getting back to you on your PRs.

And I agree that selectable text is desirable, but not at the cost of complexity or performance. I wish there was a simple flag that could be enabled via Qt but it seems not 😢

PR #1088, opened 2026-07-11
File name restarts instead of increasing

sashataras:
restarts counter and assigns new file with the lowest available integer


dail8859:
Seems like a reasonable change. Thanks!

PR #1090, opened 2026-07-12
Add Shift+Scroll horizontal scrolling

### LINKED ISSUE #742: Implement Shift+Scroll horizontal scrolling
### Description

The default system uses shift+scroll to achieve horizontal movement, but I found that in notepadnext it is achieved by alt+scroll.

### Default behavior of mainstream operating systems

- Windows: Most software uses Shift+Scroll Wheel for horizontal scrolling. Alt+Scroll Wheel usually has no default function or is used for special operations (such as adjusting canvas rotation in Photoshop).
- Linux (X11 & Wayland): Shift+Scroll Wheel is usually used for horizontal scrolling. Alt+Scroll Wheel may trigger window manager functions (such as moving windows).
- macOS: Generally, Shift+Scroll Wheel is also the default way to scroll horizontally.

Maybe we can use Shift+scroll to scroll horizontally ：）

_No response_

saberoueslati:
## Description

Fixes #742

Qt does not consistently translate Shift+Scroll into horizontal scrolling. This PR extends the existing editor wheel-event watcher to provide that behavior while preserving Ctrl+Scroll zoom and existing horizontal wheel input.

It also prevents accidental primary-selection pastes caused by the physical scroll wheel occasionally registering as a middle-button press while Shift is held.

---

## Changes

**`src/ZoomEventWatcher.cpp`**

- Extended the existing viewport event filter to handle Shift+Scroll.
- Transposes vertical angle and pixel deltas into horizontal deltas.
- Preserves events that already arrive horizontally instead of transposing them again.
- Removes the Shift modifier before forwarding the event, preventing Qt from treating it as a page-sized scrollbar movement.
- Preserves the original scroll phase, inversion state, event source, and pointing device.
- Keeps Ctrl+Scroll and Ctrl+Shift+Scroll mapped to zoom.
- Suppresses Shift+middle-button presses before they reach Scintilla, preventing accidental Linux/X11 primary-selection pastes while beginning horizontal scrolling.
- Leaves regular middle-click paste without Shift unchanged.

- Updated the existing viewport event-filter comment to document both zoom handling and Shift+Scroll translation.
- The existing watcher remains installed on every editor viewport, including editors in split views.

No new setting or preference is required, and no vendored Scintilla files are modified.

Manual test after changes below :

[Screencast from 2026-07-13 00-12-51.webm](https://github.com/user-attachments/assets/573b14b1-1847-41f1-ad5e-198905ea8d86)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -183,6 +183,8 @@ qt_add_executable(NotepadNext
 	ScintillaSorter.h ScintillaSorter.cpp
 	widgets/TabsQuickActionsBar.cpp
 	widgets/TabsQuickActionsBar.h
+	ShiftMiddleClickBlocker.h ShiftMiddleClickBlocker.cpp
+	ShiftWheelToHorizontalScrollFilter.h ShiftWheelToHorizontalScrollFilter.cpp
 )
 
 set_target_properties(NotepadNext PROPERTIES
--- a/src/ShiftMiddleClickBlocker.cpp
+++ b/src/ShiftMiddleClickBlocker.cpp
@@ -1,0 +1,42 @@
+/*
+ * This file is part of Notepad Next.
+ * Copyright 2026 Justin Dailey
+ *
+ * Notepad Next is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Notepad Next is distributed in the hope that it will be useful,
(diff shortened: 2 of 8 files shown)


dail8859:
Awesome! The shift+scrolling works great!

I pushed a refactor to this. I left the current `ZoomEventWatcher` unchanged from the original and broke out the new logic into two new classes, `ShiftMiddleClickBlocker` and `ShiftWheelToHorizontalScrollFilter`. That way each class has a single responsibility.

I think the shift+middle click is specific to linux so if you could make sure that still works for you, that would be great.

saberoueslati:
@dail8859 tested it just now, it works as intended on Linux

Issue #541, opened 2024-04-07
Opening a file when another instance is already is running sometimes fails


swiftgeek:
Opening a file when another instance is already is running sometimes fail, ending up with just raising a window and not actually opening the file (ie. `MainWindow::bringWindowToForeground()` runs in primary instance, but it's not followed by `NotepadNextApplication::openFiles(const QStringList&)` or anything else )

It might be my imagination, but the more tabs are open the higher the chances of it failing

NotepadNext 0.7 running on linux

dail8859:
What version of Linux? Is the AppImage or Flatpack?

ErickWei:
I have also encountered the same problem under MAC. When NN is already open and minimized, double-clicking another file will not wake up NN. The NN window needs to be restored manually, then the file  will be imported by NN.

swiftgeek:
Archlinux, installed from [AUR](https://aur.archlinux.org/packages/notepadnext)

@ErickWei that sounds like exact opposite to my problem

ErickWei:
Ohh it could be another issue sorry



> Opening a file when another instance is already is running sometimes fail, ending up with just raising a window and not actually opening the file (ie. `MainWindow::bringWindowToForeground()` runs in primary instance, but it's not followed by `NotepadNextApplication::openFiles(const QStringList&)` or anything else )
> 
> It might be my imagination, but the more tabs are open the higher the chances of it failing
> 
> NotepadNext 0.7 running on linux
> 
> New file opened using `NotepadNext "/home/user/dir/foo bar.txt"`

swiftgeek:
Actually if I minimize like in @ErickWei case it feels like it's failing a lot less

dail8859:
> Archlinux, installed from [AUR](https://aur.archlinux.org/packages/notepadnext)

I'm not sure who maintains that repository but I'm assuming this is a general Linux issue and not specifical to Archlinux+AUR.

> I have also encountered the same problem under MAC.

There's a few different components to all this working. And it works different depending on OS.

The first part is just forcing the window to be maximized and forcing it to the front of everything else. Most operating systems deem this as being malicious and make it difficult to do this.

The `MainWindow::bringWindowToForeground()` method probably needs more specific implementations for Linux and Mac. I've had to implement one for Windows just because Qt doesn't provide a good cross-platform way to do it that I've found.

[...]

PR #1092, opened 2026-07-15
Fix SingleApplication IPC reliability and add font style selection

blreay:
1. Fix unreliable SingleApplication IPC message delivery

The default sendMessage() timeout of 100ms is insufficient when the primary instance's event loop is temporarily blocked by session auto-save (every 60s), file-change detection, or modal dialogs. This causes the secondary instance to activate the primary's window (via instanceStarted signal) but fail to deliver the file path (via sendMessage), resulting in "window activates but file doesn't open" behavior.

- Increase sendMessage() timeout from 100ms to 1000ms per attempt
- Add retry logic (3 attempts) to handle transient event-loop blocking
- Show QMessageBox warning on total failure instead of silent exit
- Change sendInfoToPrimaryInstance() return type to bool

Ref: upstream SingleApplication issue, NotepadNext #541

2. Add font style selection to support Light/Thin font variants

Qt 6 merges font family variants (Regular, Light, Bold, etc.) under a single family name in QFontComboBox, making variants like 'Sarasa Fixed SC Nerd Font Light' invisible in the font picker.

- Replace QFontComboBox with QComboBox + style QComboBox in Preferences
- Populate families via QFontDatabase::families(), styles via QFontDatabase::styles(family) with cascading updates
- Add FontStyle setting to ApplicationSettings (persisted as style name)
- Map selected style to Scintilla weight/italic via styleSetWeight() and styleSetItalic() in EditorManager (initial setup + live updates)
- Use this->settings in lambdas for MSVC -permissive- compatibility

3. Add documentation for both changes in doc/


dail8859:
Pull requests should be small, targeted changes to implement a single feature or address a single problem.

PR #1096, opened 2026-07-24
Add XML Tools > Pretty Print XML to the Edit menu

SamJaarsma:
Adds an "XML Tools" submenu under Edit with a single "Pretty Print XML" action, reformatting the whole document equivalently to xmllint --format.

Implementation notes

Uses QXmlStreamReader/QXmlStreamWriter from Qt6::Core rather than QDomDocument, to avoid adding a Qt6::Xml dependency.
Follows the ScintillaSorter pattern for reading/writing the buffer, wrapped in UndoAction so the transform is a single undo step.
Malformed input shows a warning with line/column and leaves the buffer untouched.

Scope choices

Placed under Edit rather than View since it mutates the buffer, consistent with Convert Case and Line Operations.
Operates on the whole document rather than the selection, matching xmllint --format. Selection support would be a reasonable follow-up.

Known divergence from xmllint: QXmlStreamWriter normalises attribute quoting and always emits a UTF-8 declaration.

The macOS job passes on Qt 6.10 and fails only on 6.5 and 6.8, which suggests the AGL reference comes from Qt's own CMake config and was dropped upstream between 6.8 and 6.10. That would make this a Qt-version issue in CI rather than anything in this diff.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -49,6 +49,7 @@ qt_add_executable(NotepadNext
 	SpinBoxDelegate.h
 	TranslationManager.h
 	UndoAction.h
+	XmlFormatter.h
 	ZoomEventWatcher.h
 	resources.qrc
 	scripts.qrc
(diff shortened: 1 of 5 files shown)


SamJaarsma:
Hi @dail8859, just wanted to bump this one, it's been open for a bit. Happy to make any changes if you'd like a different approach. Let me know if there's anything blocking review and merge.

PR #1101, opened 2026-07-31
Fix Find/Replace dialog size and auto-shrink to fit screen

tanish0499:
## Summary

Fixes the Find/Replace dialog growing to an excessive width. Entering a long string in the Find or Replace combo box expanded the editable combos to fit their content, forcing the whole dialog wider — and that oversized size was persisted, so it reopened huge every time.

## Root cause

`comboFind` and `comboReplace` are `editable` with a `MinimumExpanding` policy and no `sizeAdjustPolicy`, so their size hint grew to fit the longest content. That width got saved via `saveGeometry()` and restored on the next open.

## Changes

- **`FindReplaceDialog.ui`** — set `sizeAdjustPolicy = AdjustToMinimumContentsLengthWithIcon` on both combos so their size hint no longer depends on content length. Prevents the problem going forward.
- **`FindReplaceDialog.cpp`** — in `loadSettings()`, clamp the restored geometry to the available screen size, so users who already have an oversized dialog saved in `notepadnext.ini` recover automatically.

## Testing

- Long strings in Find/Replace no longer widen the dialog.
- A pre-existing oversized geometry now opens clamped to the screen.
- Recent-item history and geometry persistence still work.

### DIFF
--- a/src/dialogs/FindReplaceDialog.cpp
+++ b/src/dialogs/FindReplaceDialog.cpp
@@ -26,6 +26,7 @@
 #include <QShortcut>
 #include <QClipboard>
 #include <QGuiApplication>
+#include <QScreen>
 
 #include "ScintillaNext.h"
 #include "MainWindow.h"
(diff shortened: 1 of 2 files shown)


dail8859:
Hi @tanish0499

I do not see this behavior in the current master branch:

> Fixes the Find/Replace dialog growing to an excessive width. Entering a long string in the Find or Replace combo box expanded the editable combos to fit their content, forcing the whole dialog wider

FindWindow

tanish0499:
Hi @dail8859 , 
Sorry for creating confusion, please try "find option" with a long string, and then close the NotepadNext and reopen it. Afterthat open search dialog, it gets stretched to fill the string in one line.
I'm attaching the videos also of both mac and windows behaviour. I will also update the PR description to pinpoint exact behaviour.

https://github.com/user-attachments/assets/ea106573-6ab6-4a0a-965f-a20ddabe28f6



https://github.com/user-attachments/assets/257d994f-aa44-46c7-8f85-e94191033eea

dail8859:
Ah ok I see that now.

Thanks!

PR #1103, opened 2026-07-31
Fix Unfold All action wired to Fold All instead of actionUnfoldAll

PCCV:
There's a problem with folding: Alt+Shift+0 does nothing, and Fold All unfolds everything instead of folding when content is already folded. There is an error on line 741, actionFoldAll should be actionUnfoldAll


dail8859:
Thank you!

PR #1104, opened 2026-08-01
Add QuickFindNext action (Ctrl+F3)

PCCV:
### Description:
This adds the QuickFindNext action, present in Notepad++ but missing in NotepadNext.

### Behavior: 
pressing Ctrl+F3 searches for the next occurrence of the word under the cursor (or the current selection, if any), without opening the Find dialog — matching Notepad++'s existing behavior.

This PR builds on top of the recently merged Fix Unfold All action (#1103).

### Changes:
Added actionQuickFindNext and wired it to the corresponding search logic
MainWindow.cpp — Modify the showFindReplaceDialog function to prevent the dialog box from opening.
MainWindow..h — change prototype of MainWindow::showFindReplaceDialog
Updated MainWindow.ui to register the new action and its shortcut (Ctrl+F3)

### DIFF
--- a/src/dialogs/MainWindow.cpp
+++ b/src/dialogs/MainWindow.cpp
@@ -421,6 +421,11 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
         }
     });
 
+    connect(ui->actionQuickFindNext, &QAction::triggered, this, [=,this]() {
+		showFindReplaceDialog(FindReplaceDialog::FIND_TAB, false);
+		FindReplaceDialog *f = findChild<FindReplaceDialog *>(QString(), Qt::FindDirectChildrenOnly);
+	});
+
     connect(ui->actionQuickFind, &QAction::triggered, this, [this]() {
         QuickFindWidget *quickFind = findChild<QuickFindWidget *>(QString(), Qt::FindDirectChildrenOnly);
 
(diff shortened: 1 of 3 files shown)


src/dialogs/MainWindow.cpp
```diff
@@ -421,6 +421,11 @@ MainWindow::MainWindow(NotepadNextApplication *app) :
         }
     });
 
+    connect(ui->actionQuickFindNext, &QAction::triggered, this, [=,this]() {
+		showFindReplaceDialog(FindReplaceDialog::FIND_TAB, false);
+		FindReplaceDialog *f = findChild<FindReplaceDialog *>(QString(), Qt::FindDirectChildrenOnly);
```
dail8859, line 426: This line is not needed

src/dialogs/MainWindow.cpp
```diff
@@ -1616,7 +1621,7 @@ void MainWindow::convertEOLs(int eolMode)
     ui->statusBar->refresh(editor);
 }
 
-void MainWindow::showFindReplaceDialog(int index)
+void MainWindow::showFindReplaceDialog(int index, bool show)
```
dail8859, line 1624: Having a function called "show X" that takes a variable to not show it is a [code smell](https://en.wikipedia.org/wiki/Code_smell) so likely something needs refactored.

src/dialogs/MainWindow.cpp
```diff
@@ -1656,6 +1661,10 @@ void MainWindow::showFindReplaceDialog(int index)
     frd->show();
     frd->raise();
     frd->activateWindow();
+	if (! show) {
+		frd->find();
+		frd->reject(); 
+	}
```
dail8859, line 1667: This hurts my eyes. The entire dialog flashes on the screen since it is being shown then hidden quickly.

dail8859 (changes requested):
This is using settings already in the dialog and messing with the search. For example if "backwards" is selected in the find dialog then the search is going backwards. Same with if regex is selected. If there is a special character it will do a regex search, then when Ctrl+F3 is pressed again it doesn't give you expected results.

For example, selecting "0.25" will find "0125" but then pressing Ctrl+F3 won't find "0.25".

PR #1106, opened 2026-08-11
Added fix for sort lines by selected text

wstagg:
When doing line operations -> sort lines ascending/descending now only the selected text is sorted

Example of three rows I want to Edit > Line Operations > Sort Lines Ascending:

### Action
5
2 (selected)
4 (selected)
1 (selected)
3

### Result:

5
1
2
4
3

### DIFF
--- a/src/ScintillaSorter.cpp
+++ b/src/ScintillaSorter.cpp
@@ -27,18 +27,33 @@ ScintillaSorter::ScintillaSorter(ScintillaNext *editor)
 void ScintillaSorter::sort(const Sorter &sorter)
 {
     const QByteArray eol = editor->eolString();
-    const QByteArray text = readEditorText();
-    QVector<QByteArrayView> lines = ByteArrayUtils::split(text, eol);
+    QByteArray text = readEditorText();
+    const QByteArray selectedText = editor->getSelText();
+    QVector<QByteArrayView> lines;
+    qsizetype textStartIndex = 0;
+
+    if (!selectedText.isEmpty())
+    {
+        textStartIndex = text.indexOf(selectedText);
+        lines = ByteArrayUtils::split(selectedText, eol);
+    }
+    else
+    {
+        lines = ByteArrayUtils::split(text, eol);
+    }
 
     sorter.sort(lines);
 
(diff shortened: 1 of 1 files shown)


src/ScintillaSorter.cpp
```diff
@@ -27,18 +27,33 @@ ScintillaSorter::ScintillaSorter(ScintillaNext *editor)
 void ScintillaSorter::sort(const Sorter &sorter)
 {
     const QByteArray eol = editor->eolString();
-    const QByteArray text = readEditorText();
-    QVector<QByteArrayView> lines = ByteArrayUtils::split(text, eol);
+    QByteArray text = readEditorText();
+    const QByteArray selectedText = editor->getSelText();
```
dail8859, line 31: Just using the selected text will give odd results if partial lines are selected. It would need to sort whole lines. From the beginning of the line the selection starts, to the end of the line the end of the selection is on.

src/ScintillaSorter.cpp
```diff
@@ -27,18 +27,33 @@ ScintillaSorter::ScintillaSorter(ScintillaNext *editor)
...
+    QVector<QByteArrayView> lines;
+    qsizetype textStartIndex = 0;
+
+    if (!selectedText.isEmpty())
+    {
+        textStartIndex = text.indexOf(selectedText);
```
dail8859, line 37: This will give incorrect results if the selectedText is in multiple location in the document.

src/ScintillaSorter.cpp
```diff
@@ -27,18 +27,33 @@ ScintillaSorter::ScintillaSorter(ScintillaNext *editor)
...
+    {
+        lines = ByteArrayUtils::split(text, eol);
+    }
 
     sorter.sort(lines);
 
     QByteArray result = ByteArrayUtils::join(lines, eol);
-    writeEditorText(result);
+
+    text.replace(textStartIndex, result.size(), result.data());
```
dail8859, line 49: There's no need to update the entire file, once the start and end location are determined then only that section of the editor's contents can be replaced.

dail8859 (changes requested):

PR #1119, opened 2026-08-29
Improve MacOS build

dail8859:


hemanpa:
The macOS jobs here all pass, so this fixes the `ld: framework 'AGL' not found`
failures on `master` (Qt 6.5/6.8 link against AGL, which was removed in the macOS 26
SDK). Just the two `windows-11-arm` jobs are failing, and I think it's a one-line fix.

Both fail in `Install Qt` with:

```
ERROR   : The packages ['qt5compat', 'qt_base'] were not found while parsing XML of package information!
```

`win64_msvc2019_arm64` isn't a published architecture for these Qt versions. Checking
against the live repositories with the same aqtinstall version the action uses (3.3.0):

```
$ aqt list-qt windows desktop --arch 6.8.3   | tr ' ' '\n' | grep arm
win64_msvc2022_arm64_cross_compiled

$ aqt list-qt windows desktop --arch 6.10.0  | tr ' ' '\n' | grep arm
win64_msvc2022_arm64_cross_compiled
```

So for `host: windows` (which is what the step resolves to for these jobs) the value
should be `win64_msvc2022_arm64_cross_compiled`, on lines 36 and 41. `qt5compat` is
available for that architecture, so the `modules:` line should be fine as-is:

```
$ aqt list-qt windows desktop --modules 6.8.3 win64_msvc2022_arm64_cross_compiled | tr ' ' '\n' | grep qt5compat
qt5compat
```

There's also a native option if you'd prefer non-cross-compiled Qt on those runners:
`host: windows_arm64` with `arch: win64_msvc2022_arm64` (also has `qt5compat`).

I've only verified the architecture strings against the repositories — I wasn't able to
run the workflow on my fork, so I can't confirm the `Configure`/`Build` steps succeed
afterwards with `cmake_arch: ARM64`.

The `codespell` failure here is the pre-existing `alot` typo on `master`, which #1124
fixes.

dail8859:
I got the MacOS stuff working better it appears and at the time took a crack at arm support. But it seems https://github.com/dail8859/NotepadNext/pull/1128 has that ironed out way better than my attempt.

So hopefully will get some time soon to take the Mac stuff from this PR and the Windows ARM stuff from https://github.com/dail8859/NotepadNext/pull/1128 and everything should be in a pretty good state.

Faq:
When planned new version to test this on mac arm?

PR #1121, opened 2026-08-31
Fix deprecated implicit `this` capture in `[=]` lambdas (C++20)

PCCV:
## Summary

Replaces implicit `this` capture (`[=]`) with explicit capture (`[=, this]`) in lambdas defined inside member functions, across 18 files. This fixes the C++20 deprecation warning:

    attention: la capture implicite de « this » via « [=] » est dépréciée en C++20 [-Wdeprecated]

## Scope

- Mechanical fix, one line per occurrence, no behavior change.
- `thirdparty/scintilla/src/Editor.cxx` has one remaining occurrence of this warning, left untouched intentionally since it's vendored third-party code, not maintained here.

## Testing

Built in Debug and Release, confirmed no more `-Wdeprecated` warnings outside of the vendored Scintilla file mentioned above.

### DIFF
--- a/src/DockedEditor.cpp
+++ b/src/DockedEditor.cpp
@@ -67,7 +67,7 @@ DockedEditor::DockedEditor(QWidget *parent) : QObject(parent)
     dockManager = new ads::CDockManager(parent);
     dockManager->setStyleSheet("");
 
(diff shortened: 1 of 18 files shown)


dail8859:
Thanks!

PR #1124, opened 2026-09-08
Fix spelling error in zoom out comment

hemanpa:
The `codespell` CI check has been failing on `master` since b4c6f51 because of a
typo (`alot`) in a comment.

Verified locally with codespell 2.4.3 using the same options as the workflow
(`--check-filenames --ignore-words-list doubleclick,msdos --skip *.lua,*.ui`): the
check fails on `master` and passes with this change. It was the only error reported
in `src/`.


dail8859:
<3

Issue #730, opened 2025-03-18
Windows ARM Support


lingyicute:
### Description

Hi maintainers! First of all, thank you for this great project. Nowadays, Windows on ARM devices like the Surface Pro X are becoming increasingly popular. I would love to use this editor natively on these devices. However, the current build seems to target x86/x64 architectures only. I was wondering if there are any plans to support ARM architecture in the near future?

dail8859:
Having ARM support would be desirable, but I have no experience or way to test/debug builds for ARM. If someone was willing to contribute the required changes to the code, project files, CI/CD, etc I'd be very willing to accept it. I wouldn't be able to offer too much help but would try where I could.

NN---:
Hopefully GitHub runner supports Windows ARM soon: https://github.com/github/roadmap/issues/1098

lingyicute:
This is great news! However, after checking the [documentation](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), Windows ARM runners are not currently available for public repositories. That said, if someone could submit a PR to address architecture-specific issues, I can immediately run the build and test it on my Windows ARM device.

dail8859:
So I possibly(?) think I have some good news.

On my PC I was able to install the arm64 libraries for visual studio, and the arm64 libraries for Qt and was able to compile the code and get an executable pretty easily. I of course cannot run the application or verify anything is correct other than I get no errors. There's also no `windeployqt` for arm, so I'm not sure the best way to even begin to test this.

I could potentially zip up the equivalent arm  dlls as the normal version has, but that would be a shot in the dark.

lingyicute:
You can send the binaries here or to my email (available on my Github profile). I'll be able to test them. As a side note: Today I noticed that NotepadNext on Flathub already has an arm64 build. You might consider creating an ARM-architecture AppImage and uploading it to the releases section. If needed, I also have a Linux arm64 device available for testing it. (Of course, Github Actions' newly launched Linux arm runners could also be used for testing, though desktop environments would require manual installation.)

dail8859:
@lingyicute Apologies for the delay.

This is a complete shot in the dark so no clue if this will even work.

[arm.zip](https://github.com/user-attachments/files/19823426/arm.zip)

lingyicute:
Sorry, I’ve been busy with exams these past few days. I just got the chance to try it out now.

[...]

PR #1128, opened 2026-09-12
feat: Add Windows ARM64 support

lingyicute:
Solve https://github.com/dail8859/NotepadNext/issues/730


matthewyang204 (approved):

matthewyang204:
Seems to work fine. Demo video:
https://drive.google.com/file/d/1Mp3U0jyQCQNYxPczVOQq86WBUijXeQfW/view?usp=sharing
For reference, I used a copy of the installer obtained from the Actions run.

dail8859:
Appreciate the PR and glad to see it is working and especially the installer!

My one caution would be related to the application's auto-updater (which is the reason for the libcrypto dependency). Right now the application is pointed to check https://github.com/dail8859/NotepadNext/blob/master/updates.json which is just the x64 installer so the ARM version would end up pulling down the x64 installer and then fail to launch it.

For the time being it may be preferrable to disable the auto-update feature on arm. I did also see that the QSimpleUpdater has a custom platform key that can be set here https://github.com/alex-spataru/QSimpleUpdater/blob/c5c136c0f79ca6423b0914a2676ae3d0e3811c5b/src/Updater.cpp#L350 and then that will force it to check for a new key rather than just "windows".

matthewyang204:
We could just have it look for a download-url-arm64 key or something instead, and I think it would just update to that instead.

lingyicute:
Implemented. 🤗

lingyicute:
> Seems to work fine. 

I definitely tested it myself before posting. By the way, you can use [this project](https://github.com/lingyicute/Action-ssh) of mine to get a Windows 11 ARM cloud desktop. It works much better than QEMU.

matthewyang204 (approved):

dail8859:
FYI due to merging #1119 for fixing the MacOS build stuff there were some merge conflicts. I incorporated these changes and (hopefully) fixed the build.yml file in #1130 now.

dail8859:
Closing in favor of #1130

PR #1130, opened 2026-09-13
Add Windows ARM64 support [initial]

dail8859:
- build.yml: add native windows-11-arm jobs (Qt 6.8/6.10, the only versions with ARM64 binaries), install Qt win64_msvc2022_arm64, use the arm64 VS environment, and upload/publish -arm64 artifacts
- PackagingWindows.cmake: only ship the OpenSSL DLLs for x64 builds; ARM64 uses Qt's Schannel TLS backend
- installer.nsi: only embed the OpenSSL DLLs when present in the package

No C++ changes are required: as far as I know, the code and all thirdparty dependencies are architecture-neutral.

CC @lingyicute


lingyicute:
Just saw this! Let me test it out.

lingyicute:
截图 2026-09-14 23-16-46

